Issue #67, opened 2021-03-07
Error: Error checking availability ZeroMatchesError: unable to find SGV1, zero matches


developer-guy:
Hello, I'm trying to demonstrate the k3s cluster setup using the k3sup tool but I'm getting the following error, what am I doing wrong here, thank you.

 ```bash
 $ civo instance create --hostname k3sup-1.demo --sshkey cf782af6-6c36-4af0-bd4b-f4c4bfdda2a0 --template 811a8dfb-8202-49ad-b1ef-1e6320b20497 --region NYC1 --size g3.small --wait
 
Error: Error checking availability ZeroMatchesError: unable to find SGV1, zero matches
```

andyjeffries:
@alejandrojnm this error message needs some work. I know what the problem is and even I don't understand that message 😂

I don't understand where SGV1 comes from, but region is specified as `NYC1` in the command

PR #72, opened 2021-03-11
Fix typo in default config

stiantoften:
The default region in the config file is set to SGV1 on first launch, but the actual region is called SVG1.

This might be the reason for the cryptic error in #67


alejandrojnm:
Yep, my bad, thanks for this

PR #86, opened 2021-05-10
Updated README to include instructions for setting region from CLI before creating a new cluster.

akshaysin:
This helps with the error you might see otherwise :

Error: Error checking availability ZeroMatchesError: unable to find SVG1, zero matches


andyjeffries:
Thank you for the PR, however I've actually just fixed this in a better way (warning outputted from the CLI with steps how to fix, don't set to an incorrect default value, set a working region when adding an API key), so there's no need to add documentation on it. Thank you for the time anyway and we appreciate the bug report.

Issue #101, opened 2021-07-07
Input api key via stdin


alexellis:
I'd like to be able to input the api key via stdin, not just via an argument to the command.

Why? It exposes a confidential token in my bash history and could leak or be sent back by analytics tools or exposed on a live stream when finding past commands.

OpenFaaS' CLI enables the use of a file i.e. `cat password | faas-cli login` or `echo $PASSWORD | faas-cli login`

You're welcome to copy the code / approach that we use in the login command.

I'd feel more comfortable using the Civo CLI if this change was made to increase security.

https://github.com/openfaas/faas-cli/blob/master/commands/login.go

andyjeffries:
Good suggestion but there's nothing wrong with using $PASSWORD (or $API_KEY) with the Civo CLI instead of the raw API key too. I do think asking for it interactively if not supplied as a parameter is a good idea though.

andyjeffries:
@RealHarshThakur this might be a nice easy one to get you in to the CLI codebase.

RealHarshThakur:
Added this feature, so closing this. :)

PR #102, opened 2021-07-08
Add api key through stdin

RealHarshThakur:
Ref:  https://github.com/civo/cli/issues/101

### DIFF
--- a/cmd/apikey_save.go
+++ b/cmd/apikey_save.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"bufio"
 	"fmt"
 	"os"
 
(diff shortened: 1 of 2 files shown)


andyjeffries:
Unless I'm misunderstanding, this PR changes it so that you can ONLY enter them via STDIN? That will break CI usage.

> I do think asking for it interactively _if not supplied as a parameter_ is a good idea though
https://github.com/civo/cli/issues/101#issuecomment-875535318

Ignore this ^ see the change further down.

cmd/apikey_save.go
```diff
@@ -1,26 +1,51 @@
...
 
+		var username, apikey string
+		var err error
+		if len(args) == 0 {
+			reader := bufio.NewReader(os.Stdin)
+			fmt.Printf("Enter Username: ")
+
+			username, err = reader.ReadString('\n')
+			if err != nil {
+				utility.Error("Error reading username")
```
andyjeffries: Feels like maybe we should give the customer more to go on? Does the error have a string version?

cmd/apikey_save.go
```diff
@@ -1,26 +1,51 @@
...
+	Args:    cobra.MinimumNArgs(0),
 	Example: "civo apikey save NAME KEY",
 	Run: func(cmd *cobra.Command, args []string) {
-		config.Current.APIKeys[args[0]] = args[1]
 
+		var username, apikey string
+		var err error
+		if len(args) == 0 {
+			reader := bufio.NewReader(os.Stdin)
+			fmt.Printf("Enter Username: ")
```
andyjeffries: This should be "Enter a nice name for this account/API Key:" - it's not their username.

cmd/apikey_save.go
```diff
@@ -1,26 +1,51 @@
...
 	Use:     "save",
 	Aliases: []string{"add", "store", "create", "new"},
 	Short:   "Save a new API key",
-	Args:    cobra.MinimumNArgs(2),
+	Args:    cobra.MinimumNArgs(0),
 	Example: "civo apikey save NAME KEY",
 	Run: func(cmd *cobra.Command, args []string) {
-		config.Current.APIKeys[args[0]] = args[1]
 
+		var username, apikey string
```
andyjeffries: `name` would be better than `username`, see the comment below.

Also Go variables should be camelCase so `apiKey`.

cmd/apikey_save.go
```diff
@@ -1,26 +1,51 @@
...
+		if len(args) == 0 {
+			reader := bufio.NewReader(os.Stdin)
+			fmt.Printf("Enter Username: ")
+
+			username, err = reader.ReadString('\n')
+			if err != nil {
+				utility.Error("Error reading username")
+				os.Exit(1)
+			}
+			fmt.Printf("Now, please type in the apikey: ")
```
andyjeffries: "Enter the API key:"

(no need for "Now, ", keeping the language consistent with the other prompt - you used "enter" not "type", and it was in an instructional format rather than request, i.e. no "please")

cmd/apikey_save.go
```diff
@@ -1,26 +1,51 @@
...
 	Use:     "save",
 	Aliases: []string{"add", "store", "create", "new"},
 	Short:   "Save a new API key",
-	Args:    cobra.MinimumNArgs(2),
+	Args:    cobra.MinimumNArgs(0),
 	Example: "civo apikey save NAME KEY",
 	Run: func(cmd *cobra.Command, args []string) {
-		config.Current.APIKeys[args[0]] = args[1]
 
+		var username, apiKey string
```
andyjeffries: not `username`, just `name`. Remember, it's not their username but a key.

`civo apikey save dev hodihsofidjsoifjdsoijf`

You can use whatever you want as the name, it's just a reference for when listing them in the CLI, it isn't used anywhere.

andyjeffries:
Last fix needed, then it's all good - merge, update changelog, bump version (tag commit) and it should be automatically built and released.

PR #119, opened 2021-08-20
Add '--firewall' flag and validation to instance create command

### LINKED ISSUE #116: Instance created in non-default network gets default firewall
Screenshot 2021-08-20 at 11 20 10 AM

This shouldn't be. What we should do is:

* If user creates an instance in non-default network but didn't provide `--firewall` flag, we should raise an error and ask the user which firewall (from selected non-default network) to use
  * If firewall exists, proceed
  * If firewall not exist, raise another error

zulh-civo:
In order to merge this PR, civogo needs to add `FirewallID` in the [`InstanceConfig`](https://github.com/civo/civogo/blob/master/instance.go#L68) struct.

When that's done, I'll proceed with:

- [x] ~~Publishing new civogo release~~
- [x] ~~Update it here in the [go.mod](https://github.com/civo/cli/blob/master/go.mod) file~~
- [x] ~~Change this PR status from WIP/Draft to Open~~

Closes #116


zulh-civo:
Upstream PR -> https://github.com/civo/civogo/pull/59

alejandrojnm (approved):

PR #128, opened 2021-09-02
Remove "move IP" from "civo instance create" command and docs

### LINKED ISSUE #127: Remove "move from" IP in "civo instance create" command
It's legacy feature and we should remove it.

Current:

```
-p, --publicip move_ip_from:intances_id   This should be either none, create or move_ip_from:intances_id by default is create (default "create")
```

Checklist:

- [x] Remove from the code
- [x] Update the docs

zulh-civo:
Close #127


alejandrojnm (approved):

iambenmitchell:
How do we move IPs between instances? Since we cannot rebuild an instance

PR #132, opened 2021-09-22
Disable "civo k3s upgrade" command

### LINKED ISSUE #131: Disable "civo k3s upgrade" command
Since there's no other k8s version to upgrade (as of today), we decided to disable this command first. When we have more available versions later, we will re-enable this.

zulh-civo:
Close #131


saiyam1814 (approved):
LGTM

PR #152, opened 2021-09-29
Add support to set API key from environment variables

### LINKED ISSUE #107: Add support for passing CIVO_TOKEN
For more convenience, it would be cool to be able also to set the API key and region via an environment variable. This would allow it easier to use in CI Systems for example GitLab CI.

Currently, I use a short workaround like this:

```shell
 civo apikey add civo-k3s $CIVO_TOKEN
 civo apikey current civo-k3s
 civo kubernetes config $(cat terraform/civo-cloud/CLUSTER_NAME) --save --merge
```
 I found that the terraform provider already supports this behavior. I think it would be also a great extension to the CLI.

zulh-civo:
Close #107 

---

From https://github.com/civo/cli/issues/107#issuecomment-929839953:

> When it's merged and released, it will support this flow:
> 
> ```
> $ export CIVO_API_KEY_NAME=personal
> $ export CIVO_API_KEY=<CIVO-API-KEY>
> $ civo apikey save --load-from-env
> ```
> 
> Which in turns will create/update _~/.civo.json_ file to:
> 
> ```json
> {
>  "apikeys": {
>     "personal": "<CIVO-API-KEY>"
>   },
>   "meta": {
>     "admin": false,
>     "current_apikey": "personal",
>     "default_region": "FRA1",
>     "latest_release_check": "2021-09-29T12:45:35+08:00",
>     "url": "https://api.civo.com"
>   }
> }
> ```

### DIFF
--- a/cmd/apikey.go
+++ b/cmd/apikey.go
@@ -47,4 +47,7 @@ func init() {
 	apikeyCmd.AddCommand(apikeyRemoveCmd)
 	apikeyCmd.AddCommand(apikeyCurrentCmd)
 	apikeyCmd.AddCommand(apikeyShowCmd)
+
+	// Flags for "civo apikey save" command
+	apikeySaveCmd.Flags().BoolVar(&loadApiKeyFromEnv, "load-from-env", false, "When set, the key will be loaded from $CIVO_API_KEY_NAME and $CIVO_API_KEY")
 }
--- a/cmd/apikey.go
+++ b/cmd/apikey.go
@@ -49,5 +49,5 @@ func init() {
 	apikeyCmd.AddCommand(apikeyShowCmd)
 
 	// Flags for "civo apikey save" command
-	apikeySaveCmd.Flags().BoolVar(&loadApiKeyFromEnv, "load-from-env", false, "When set, the key will be loaded from $CIVO_API_KEY_NAME and $CIVO_API_KEY")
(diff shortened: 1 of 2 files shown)


rberrelleza:
This is great! 


My only comment is with regards to having the API Key Name mandatory.  AFAIK I can only create one API key in my civo account, so what's the purpose of me providing a name? Maybe we could make that optional, and default to something like 'civo-api', or the hostname of the machine?

zulh-civo:
@rberrelleza thanks for your feedback, Ramiro. That's fair point. I was considering people with multiple Civo accounts (e.g. one for personal, one for work).

But, after carefully think about your comment above, I think it's a valid concern and we should only make the API key required. When no name is specified, it will fallback to the hostname where the CLI is running.

People with multiple accounts can re-set the environment variables and re-run the command like this:

```
$ export CIVO_API_KEY_NAME=account-1
$ export CIVO_API_KEY=<CIVO-API-KEY-1>
$ civo apikey save --load-from-env

$ export CIVO_API_KEY_NAME=account-2
$ export CIVO_API_KEY=<CIVO-API-KEY-2>
$ civo apikey save --load-from-env
```

I've updated the code and one of my co-worker will review this soon. You can expect this feature in our next release. Thank you again for sharing your feedback with us.

alejandrojnm (commented):
LGTM

PR #162, opened 2021-11-02
Disable load balancer files

zulh-civo:
Because we don't have Load Balancer feature (yet) in [CivoStack](https://www.civo.com/blog/new-year-new-york-new-civostack)


alejandrojnm (approved):
LGTM

PR #178, opened 2022-01-24
Enable load balancer commands

vishalanarase:
Enable load balancer `list, show` command
 Add quota fields for load balancer


alejandrojnm:
@vishalanarase I found a lot of `lb` like `lbNetwork` I think is better `loadBalancerNetwork` Explicit is better than implicit.

alejandrojnm (approved):
LGTM

PR #184, opened 2022-02-14
Update application install logic

haardikdharma10:
## Summary

This PR aims to modify CLI code in such a way that `civo k3s create -r metrics-server, traefik-v2-nodeport` would send an empty string in applications field.

## Background

NA

### DIFF
--- a/cmd/kubernetes_create.go
+++ b/cmd/kubernetes_create.go
@@ -201,6 +201,9 @@ var kubernetesCreateCmd = &cobra.Command{
 					utility.Warning("the app that tries to install %s is not valid", v)
 					os.Exit(1)
 				}
+				if utility.CheckIfAppHasPrefix(v) {
+					break
+				}
 			}
 			configKubernetes.Applications = installApplications
 		}
--- a/cmd/kubernetes_create.go
+++ b/cmd/kubernetes_create.go
@@ -175,6 +175,9 @@ var kubernetesCreateCmd = &cobra.Command{
 			for _, v := range strings.Split(removeapplications, ",") {
 				if utility.CheckAPPName(v) {
 					rmApp = append(rmApp, fmt.Sprintf("-%s", v))
+					if utility.Contains(rmApp, "-metrics-server") || utility.Contains(rmApp, "-traefik-v2-nodeport") {
+						continue
+					}
 				} else {
 					utility.Warning("the app that tries to remove %s is not valid", v)
 					os.Exit(1)
@@ -200,9 +203,6 @@ var kubernetesCreateCmd = &cobra.Command{
 				if !utility.CheckAPPName(v) {
 					utility.Warning("the app that tries to install %s is not valid", v)
 					os.Exit(1)
-				}
-				if utility.CheckIfAppHasPrefix(v) {
-					break
 				}
 			}
 			configKubernetes.Applications = installApplications
--- a/cmd/kubernetes_create.go
+++ b/cmd/kubernetes_create.go
@@ -16,7 +16,7 @@ import (
 
 var numTargetNodes int
 var waitKubernetes, saveConfigKubernetes, mergeConfigKubernetes, switchConfigKubernetes bool
-var kubernetesVersion, targetNodesSize, clusterName, applications, removeapplications, installApplications, networkID, existingFirewall, createFirewall, cniPlugin string
+var kubernetesVersion, targetNodesSize, clusterName, applications, removeapplications, networkID, existingFirewall, createFirewall, cniPlugin string
 var kubernetesCluster *civogo.KubernetesCluster
 
 var kubernetesCreateCmdExample = `civo kubernetes create CLUSTER_NAME [flags]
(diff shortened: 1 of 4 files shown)


RealHarshThakur:
No background/description?

RealHarshThakur:
I just noticed there's a `-r`  flag on the CLI to mention which apps to not include. So I think we should just use that. 

We need to modify code in such a way that
> civo k3s create -r metrics-server, traefik-v2-nodeport

would send an empty string in applications field.

haardikdharma10:
Thanks @RealHarshThakur, updated the PR.

RealHarshThakur:
We need to communicate this change to the user too. So it needs some examples/description in the commands

cmd/kubernetes_create.go
```diff
@@ -175,6 +175,9 @@ var kubernetesCreateCmd = &cobra.Command{
 			for _, v := range strings.Split(removeapplications, ",") {
 				if utility.CheckAPPName(v) {
 					rmApp = append(rmApp, fmt.Sprintf("-%s", v))
+					if utility.Contains(rmApp, "-metrics-server") || utility.Contains(rmApp, "-traefik-v2-nodeport") {
```
RealHarshThakur: Are we looping over the same list twice? Can we simply not compare `v` with metrics-server and traefik?

RealHarshThakur (changes requested):

cmd/kubernetes_create.go
```diff
@@ -175,6 +175,9 @@ var kubernetesCreateCmd = &cobra.Command{
 			for _, v := range strings.Split(removeapplications, ",") {
 				if utility.CheckAPPName(v) {
 					rmApp = append(rmApp, fmt.Sprintf("-%s", v))
```
RealHarshThakur: Sending `-` in applications field won't be a thing anymore in the API, so we should probably be removing this too.

RealHarshThakur:
Couple commands you'll need to ensure are working as part of this PR:

`civo k3s create -a foo` => applications=metrics-server,foo,traefik-v2-nodeport
`civo k3s create -a foo -r metrics-server` => applications=traefik-v2-nodeport,foo

cmd/kubernetes_create.go
```diff
@@ -175,6 +175,9 @@ var kubernetesCreateCmd = &cobra.Command{
...
+					if utility.Contains(rmApp, "-metrics-server") || utility.Contains(rmApp, "-traefik-v2-nodeport") {
+						continue
+					}
 				} else {
 					utility.Warning("the app that tries to remove %s is not valid", v)
 					os.Exit(1)
```
RealHarshThakur: You'll need to modify the installApplications field pretty much always too so that we explicitly add metrics server and traefik.

haardikdharma10:
> I just noticed there's a `-r` flag on the CLI to mention which apps to not include. So I think we should just use that.
> 
> We need to modify code in such a way that
> 
> > civo k3s create -r metrics-server, traefik-v2-nodeport
> 
> would send an empty string in applications field.

But the user can do something like `civo k3s create -a foo,-traefik-v2-nodeport` instead of using the `-r` flag, right? or no? which will return applications=metrics-server,foo 

I believe we should keep both the options for the user to decide. If the user feels like typing out another flag is inconvenient, he/she can just add `-` in the -a flag.

RealHarshThakur:
Personally, I don't find using `-`  intuitive to exclude something when passing a list but that could just be me. 
Especially, considering apps can only be removed during creating a cluster.

@andyjeffries  WDYT?

andyjeffries:
I'm happy having it officially documented as using `-r` but I believe either that or `-` works currently, so we can't break existing customers' usage for such a minor improvement.

RealHarshThakur:
Skimming the code, I don't think we support it right now. It would be a new thing we'll introduce

andyjeffries:
If that's the case, I'm happy not adding it. However... I would suggest testing the current CLI against the current API and seeing how it works, rather than a theoretical skim of the code and not finding it.

haardikdharma10:
Confirmed that something like `civo k3s create -a foo,-traefik-v2-nodeport` doesn't work. It gives the error - `the app that tries to install -traefik-v2-nodeport is not valid`. So i think we'll go ahead with the `-r` plan and then document it.

cmd/kubernetes_test.go
```diff
@@ -0,0 +1,39 @@
...
+		{
+			name: "Test Removal of Default Apps",
+			args: InstallApplicationsArgs{defaultApps: []string{"metrics-server", "traefik"}, apps: "", removeApps: "traefik"},
+			want: []string{"metrics-server"},
+		},
+		{
+			name: "Test Installation of Default Apps and other Apps",
+			args: InstallApplicationsArgs{defaultApps: []string{"metrics-server", "traefik"}, apps: "foo", removeApps: ""},
+			want: []string{"metrics-server", "traefik", "foo"},
+		},
```
RealHarshThakur, line 29: Test case for :

 `InstallApplicationsArgs{defaultApps: []string{"metrics-server", "traefik"}, apps: "foo", removeApps: "traefik"}`

RealHarshThakur (dismissed):

RealHarshThakur:
Docs/help info is still missing.

haardikdharma10:
Added 👍

RealHarshThakur (approved):

PR #187, opened 2022-02-21
Modify the create-firewall flags

alejandrojnm:
Modify the way to create the firewall from the Kubernetes command


README.md
```diff
@@ -888,10 +888,10 @@ To create a new Firewall, use `civo firewall create new_firewall_name`:
 $ civo firewall create civocli_demo
 Created a firewall called civocli_demo with ID ab2a25d7-edd4-4ecd-95c4-58cb6bc402de
 ```
-Also you can create a firewall without rules by default, using the flag `-r` or `--createrules` and set to `false`, in both cases you need use like this
+Also you can create a firewall without rules by default, using the flag `-r` or `--create-rules` and set to `false`, in both cases you need use like this
```
kaihoffman: I would fix this to be grammatically as follows:

```
You can also create a firewall without any default rules by using the flag `-r` or `--create-rules` set to `false`. In both cases, the usage is like:
```

cmd/kubernetes.go
```diff
@@ -85,7 +85,8 @@ func init() {
 	kubernetesCreateCmd.Flags().BoolVarP(&mergeConfigKubernetes, "merge", "m", false, "merge the config with existing kubeconfig if it already exists.")
 	kubernetesCreateCmd.Flags().BoolVarP(&switchConfigKubernetes, "switch", "", false, "switch context to newly-created cluster")
 	kubernetesCreateCmd.Flags().StringVarP(&existingFirewall, "existing-firewall", "e", "", "optional, ID of existing firewall to use")
-	kubernetesCreateCmd.Flags().StringVarP(&createFirewall, "create-firewall", "c", "", "optional, semicolon-separated list of ports to open - leave blank for default (80;443;6443) or you can use \"all\"")
+	kubernetesCreateCmd.Flags().StringVarP(&rulesFirewall, "firewall-rules", "u", "", "optional, can be used if the --create-firewall flag is set, semicolon-separated list of ports to open")
+	kubernetesCreateCmd.Flags().BoolVarP(&createFirewall, "create-firewall", "c", false, "optional, create a firewall for the cluster with all open ports")
```
kaihoffman, line 89: Is this the desired behaviour? I thought we're opening 80, 443, 6443 by default for Kubernetes firewalls?

cmd/kubernetes_create.go
```diff
@@ -15,15 +15,17 @@ import (
...
-var waitKubernetes, saveConfigKubernetes, mergeConfigKubernetes, switchConfigKubernetes bool
-var kubernetesVersion, targetNodesSize, clusterName, applications, removeapplications, networkID, existingFirewall, createFirewall, cniPlugin string
+var waitKubernetes, saveConfigKubernetes, mergeConfigKubernetes, switchConfigKubernetes, createFirewall bool
+var kubernetesVersion, targetNodesSize, clusterName, applications, removeapplications, networkID, existingFirewall, cniPlugin, rulesFirewall string
 var kubernetesCluster *civogo.KubernetesCluster
 
 var kubernetesCreateCmdExample = `civo kubernetes create CLUSTER_NAME [flags]
 
 Notes:
+* The '--create-firewall' It will be create the default rules for the firewall.
```
kaihoffman: Above it says that creating a firewall will open all ports, and here "default" rules? Are they the same thing?

Grammatically, I suggest

`* The '--create-firewall' will add add default rules for the firewall if none are specified.`

That's assuming the "default rules" are defined somewhere.

cmd/kubernetes_create.go
```diff
@@ -33,9 +35,7 @@ Notes:
 * So the following would all be valid:
     * "80,443,6443:0.0.0.0/0;8080:1.2.3.4" (open 80,443,6443 to public and 8080 just for 1.2.3.4/32)
     * "80,443,6443;6000-6500:4.4.4.4/24" (open 80,443,6443 to public and 6000 to 6500 just for 4.4.4.4/24)
-* When '--create-firewall' flag is blank, your cluster will be created with the following rules:
-    * "80;443;6443" (open 80,443,6443 to public)
-* To open all ports for public access, "all" can be provided to '--create-firewall' flag (not recommended)
+* When '--create-firewall' flag is blank, your cluster will be created with the default rules all open
```
kaihoffman: see comments above. We seem to be using "default" and "all open" and somewhere even 80, 443, 6443 - Should be consistent throughout.

alejandrojnm: I remove this line because the first one say the same:
`* The '--create-firewall' will add default rules for the firewall if none are specified by using '--firewall-rules'.`

cmd/kubernetes.go
alejandrojnm: mm, I will say the user needs to write these ports using the `--firewall-rules` flag

cmd/kubernetes_create.go
```diff
@@ -22,20 +22,19 @@ var kubernetesCluster *civogo.KubernetesCluster
 var kubernetesCreateCmdExample = `civo kubernetes create CLUSTER_NAME [flags]
 
 Notes:
-* The '--create-firewall' It will be create the default rules for the firewall.
+* The '--create-firewall' will add default rules for the firewall if none are specified by using '--firewall-rules'.
```
kaihoffman: OK, so if you are changing the flag to require a list of ports, this can be confusing to the user. What are the default rules in this case?

alejandrojnm: all port open

kaihoffman: Yeah, I think that needs to be clear in this (and all cases) - especially since there is a part of the code where it says doing "all" is _not_ recommended.

RealHarshThakur (approved):

PR #189, opened 2022-02-23
Fix incorrect command for upgrades

haardikdharma10:
### Summary

`civo k3s cluster show <CLUSTER_NAME>` showed incorrect command for upgrades. This PR aims to fix that.

### DIFF
--- a/cmd/kubernetes_show.go
+++ b/cmd/kubernetes_show.go
@@ -11,6 +11,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var stableVersion = "1.22.2"
+
 var kubernetesShowCmd = &cobra.Command{
 	Use:     "show",
 	Aliases: []string{"get", "inspect"},
@@ -97,9 +99,11 @@ If you wish to use a custom format, the available fields are:
 			ow.WriteKeyValues()
 
 			if kubernetesCluster.UpgradeAvailableTo != "" {
-				fmt.Println()
-				fmt.Printf(utility.Red("* An upgrade to v%s is available. To upgrade use: civo k3s upgrade %s --version %s"), kubernetesCluster.UpgradeAvailableTo, kubernetesCluster.Name, kubernetesCluster.UpgradeAvailableTo)
-				fmt.Println()
+				if kubernetesCluster.Version < stableVersion {
+					fmt.Println()
+					fmt.Printf(utility.Red("* An upgrade to v%s-k3s1 is available. Learn more about how to upgrade: civo k3s upgrade --help"), stableVersion)
+					fmt.Println()
+				}
 			}
 
 			if len(kubernetesCluster.Instances) > 0 {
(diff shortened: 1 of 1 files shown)


cmd/kubernetes_show.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var stableVersion = "1.22.2"
```
RealHarshThakur: Rather than using a hardcoded version, can you fetch all the versions and check which one is stable in there? Probably check what `civo k3s versions` is doing because it already outputs non-deprecated versions.

cmd/kubernetes_show.go
```diff
@@ -97,9 +99,11 @@ If you wish to use a custom format, the available fields are:
 			ow.WriteKeyValues()
 
 			if kubernetesCluster.UpgradeAvailableTo != "" {
-				fmt.Println()
-				fmt.Printf(utility.Red("* An upgrade to v%s is available. To upgrade use: civo k3s upgrade %s --version %s"), kubernetesCluster.UpgradeAvailableTo, kubernetesCluster.Name, kubernetesCluster.UpgradeAvailableTo)
-				fmt.Println()
+				if kubernetesCluster.Version < stableVersion {
```
RealHarshThakur: This wouldn't show the development versions we'll have in the future. Any version above stable version would be a development version.

RealHarshThakur (changes requested):

cmd/kubernetes_show.go
haardikdharma10: `civo k3s versions` only outputs this - 

```
+-------------+--------+---------+
| Version     | Type   | Default |
+-------------+--------+---------+
| 1.22.2-k3s1 | stable | true    |
+-------------+--------+---------+
```

cmd/kubernetes_show.go
```diff
@@ -96,53 +96,65 @@ If you wish to use a custom format, the available fields are:
...
+				os.Exit(1)
 			}
+			if kubernetesCluster.UpgradeAvailableTo != "" {
+				for _, version := range kubeVersions {
+					if version.Type == "deprecated" {
+						continue
+					}
+					if version.Version > kubernetesCluster.Version {
+						fmt.Println()
+						fmt.Printf(utility.Red("* An upgrade to v%s-k3s1 is available. Learn more about how to upgrade: civo k3s upgrade --help"), version.Version)
```
RealHarshThakur: We shouldn't be printing this line multiple times. We should create a single string with all the versions to display and then print it out. 

Also, this code is a lot similar to what we're doing in this file, so maybe create a common function and invoke it at both places? https://github.com/civo/cli/blob/master/cmd/kubernetes_list_version.go

haardikdharma10: We can't really create a common function, some bits are different in `kubernetes_list_version.go`. For example, here we list the available versions, check if they're deprecated, and if not all the versions with `stable` and `development` are available for upgrade. While there, we we list the available versions, check if they're deprecated, and then list them using the output writer.

haardikdharma10: I have solved what you mentioned above regarding creating a single string with all the versions to display and then printing it out.

RealHarshThakur (approved):

PR #190, opened 2022-02-23
Add the ssh_key_id to the instance show

alejandrojnm:


cmd/instance_show.go
```diff
@@ -76,6 +77,7 @@ If you wish to use a custom format, the available fields are:
 		ow.AppendDataWithLabel("initial_user", instance.InitialUser, "Initial User")
 		ow.AppendDataWithLabel("initial_password", instance.InitialPassword, "Initial Password")
 		ow.AppendDataWithLabel("ssh_key", instance.SSHKey, "SSH Key")
+		ow.AppendDataWithLabel("ssh_key_id", instance.SSHKeyID, "SSH Key ID")
```
haardikdharma10, line 79: why do we need `ssh_key_id` if we already have `ssh_key`?

alejandrojnm: I think we don't need `ssh_key ` just the `ssh_key_id`

haardikdharma10 (approved):

PR #193, opened 2022-03-05
Fixed error in the save sub cmd

alejandrojnm:
Fixed error in the save sub cmd in the apikey cmd, also we add info to that cmd saying that cmd did not need arguments

Signed-off-by: Alejandro J. Nuñez Madrazo <alejandrojnm@gmail.com>


alejandrojnm:
Auto approval, was just and small change

Issue #196, opened 2022-03-07
Documenting and using civo config env variable


kameshsampath:
Currently the CIVO cli uses the env variable `CIVOCONFIG` to use the file location of `civo.json`, though its used internally its not documented.

haardikdharma10:
Closing this as this is done. Thanks @kameshsampath!

PR #198, opened 2022-03-07
Update env var

### LINKED ISSUE #197: Using the name CIVO_CONFIG for env variable over CIVOCONFIG
It will be good follow the ENV variable naming convention of having `CIVO_CONFIG` over `CIVOCONFIG`.  
And update the documentation as needed https://github.com/civo/cli/issues/196

kameshsampath:
Rename the Civo config environment variable and document its usage

Fix: #197 , #196

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -32,7 +32,7 @@ var Filename string
 
 // ReadConfig reads in config file and ENV variables if set.
 func ReadConfig() {
-	filename, found := os.LookupEnv("CIVOCONFIG")
+	filename, found := os.LookupEnv("CIVO_CONFIG")
 	if found {
 		Filename = filename
 	}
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -32,7 +32,7 @@ var Filename string
 
 // ReadConfig reads in config file and ENV variables if set.
 func ReadConfig() {
-	filename, found := os.LookupEnv("CIVOCONFIG")
+	filename, found := os.LookupEnv("CIVO_CONFIG")
```
RealHarshThakur, line 35: I do think this is better but little worried that this could be a breaking change if someone already relied on it.

kameshsampath: @RealHarshThakur  thats fair point but I am wondering if anyones is using it as until someone digs the code the will not know if such an env variable exists. Thats why I added another issue to update the README with usage. WDYT?

RealHarshThakur: Yep, agreed it's unlikely since it wasn't documented.

RealHarshThakur (approved):

Issue #200, opened 2022-03-08
Bug: `civo k8s create cluster_name` no longer works


kaihoffman:
I used to be able to create a default cluster (3 nodes, etc) by running `civo k8s create clustername`

Now it gives me:

```console
civo k8s create knative-test
Error: You can't use --firewall-rules without --create-firewall flag
```

It works if I provide the `--create-firewall` switch but this is a breaking change from a number of tutorials etc.

andyjeffries:
In a similar vein, the quick create for an instance doesn't work either:

```
$ civo instance create dokku-demo
Error: required flag(s) "diskimage", "size" not set
```

It should default to creating a medium latest Ubuntu instance. If I specify those fields I still get some weirdness...

```
$ civo instance create dokku-demo -t ubuntu-focal -i g3.medium
The instance tiny-oak has been created
```

What!? `tiny-oak`?? What happened to my choice of `dokku-demo`? Apparently this is documented, but unexpected when the K3s command Kai showed just uses it as the first positional parameter and `civo instance show tiny-oak` works as expected. I'd suggest allowing either value - first un-switched parameter if present, if not, then the `-n` flag.

alejandrojnm:
@haardikdharma10 this is done ??

haardikdharma10:
> @haardikdharma10 this is done ??

yes, only cutting a tag is left 😃

haardikdharma10:
Fixed in #203. Closing.

PR #203, opened 2022-03-10
Change size and diskimage flags as "not required"

haardikdharma10:
### Background

https://github.com/civo/cli/issues/200

### DIFF
--- a/cmd/instance_create.go
+++ b/cmd/instance_create.go
@@ -16,34 +16,13 @@ import (
 )
 
 var wait bool
-var hostnameCreate, size, diskimage, publicip, initialuser, sshkey, tags, network, firewall string
+var size, diskimage, publicip, initialuser, sshkey, tags, network, firewall string
 
 var instanceCreateCmd = &cobra.Command{
 	Use:     "create",
-	Aliases: []string{"new"},
+	Aliases: []string{"new", "add"},
 	Short:   "Create a new instance",
-	Example: "civo instance create --hostname=foo.example.com [flags]",
-	Long: `You can create an instance with a hostname parameter, as well as any options you provide.
-If you wish to use a custom format, the available fields are:
-
-	* id
-	* hostname
-	* size
-	* region
-	* public_ip
-	* status
-	* network_id
-	* diskimage_id
-	* initial_user
-	* ssh_key
-	* notes
-	* firewall_id
-	* tags
-	* script
-	* created_at
-	* reverse_dns
-	* private_ip
-	* public_ip`,
+	Example: "civo instance create HOSTNAME [flags]",
 	Run: func(cmd *cobra.Command, args []string) {
 		utility.EnsureCurrentRegion()
 
@@ -73,12 +52,12 @@ If you wish to use a custom format, the available fields are:
 			os.Exit(1)
 		}
 
-		if hostnameCreate != "" {
-			if utility.ValidNameLength(hostnameCreate) {
+		if len(args) > 0 {
+			if utility.ValidNameLength(args[0]) {
 				utility.Warning("the hostname cannot be longer than 63 characters")
 				os.Exit(1)
 			}
-			config.Hostname = hostnameCreate
+			config.Hostname = args[0]
 		}
 
 		if region != "" {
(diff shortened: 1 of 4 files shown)


cmd/instance_create.go
```diff
@@ -73,12 +52,12 @@ If you wish to use a custom format, the available fields are:
 			os.Exit(1)
 		}
 
-		if hostnameCreate != "" {
-			if utility.ValidNameLength(hostnameCreate) {
+		if len(args) > 0 {
```
alejandrojnm, line 84: I think we can change this to use https://github.com/civo/cli/blob/fa542ad26368fe8fe797f4057912a1e655525b82/utility/random.go#L44 in the case of the user don't pass the hostname

cmd/instance_create.go
```diff
@@ -16,34 +16,13 @@ import (
...
-	* ssh_key
-	* notes
-	* firewall_id
-	* tags
-	* script
-	* created_at
-	* reverse_dns
-	* private_ip
-	* public_ip`,
+	Example: "civo instance create HOSTNAME [flags]",
```
alejandrojnm, line 25: Why we remove the available fields??

cmd/instance_create.go
haardikdharma10: @alejandrojnm if the user doesn't pass the hostname it will be automatically generated from - https://github.com/civo/civogo/blob/master/instance.go#L176 

it is tested and verified 👍

cmd/instance_create.go
haardikdharma10: Added

andyjeffries:
Hold on, this doesn't implement the bug fix required. Specifically this is removing the flag and changing it to be the correct/expected behaviour. However, people may have scripts that rely on the flag - so BOTH mechanisms need to be supported.

haardikdharma10:
I am not removing the flags for size and diskimages. I am just marking them as "not required". Did I misunderstood something?

andyjeffries:
This comment only relates to this merge request. This MR removes the hostname flag. It shouldn't remove it but accept EITHER the hostname flag or the first positional parameter. These should be equivalent:

```
civo instance create my-demo
civo instance create -s my-demo
civo instance create --hostname=my-demo
```

andyjeffries:
(the stuff about size, disk images is in a separate MR - or missing as it stands)

andyjeffries:
Ohhh, I see the title of the MR is one thing, but in the code you do things with disk image and size too. OK, I'll address those:

You seem to have removed the "required" part of the CLI spec. However, if you don't specify a disk image it will error as the code stands, correct? That's not right behaviour. It should have a default disk image in there.

In short, this command should work - can you test it your side please:

```
civo instance create my-demo
```

No specification of size (medium is OK) nor disk image (Ubuntu 20.04 should be default - for now)

In fact, even this should work:

```
civo instance create
```

(it should generate a random hostname, I think it already does this though)

haardikdharma10:
Okay, so I should have changed the PR title. This MR indeed removes the hostname flag and adds it as an argument. This needs to be changed as you mentioned. The stuff about size and disk images is also in this MR. The error when creating instance with `civo instance create dokku-demo` was `Error: required flag(s) "diskimage", "size" not set` 

This was because they were marked as required flags.

Ahh I see your recent comment, all the commands that you have mentioned have been tested by me :) 

To summarize - 

- `civo instance create my-demo` creates an instance with hostname `my-demo` size `g3.medium` and disk image `Ubuntu 20.04`
-  `civo instance create` creates an instance with a random hostname size `g3.medium` and disk image `Ubuntu 20.04`
- `civo instance create new-instance -t ubuntu-focal -i g3.medium` creates an instance with hostname `new-instance` and the mentioned size and disk image.

andyjeffries:
Cool, thank you - the last part is key for me.

andyjeffries:
I am a bit lost though - where is Ubuntu 20.04 or focal specified as the default value? I can't find it on https://github.com/civo/cli/pull/203/files and the API code doesn't seem to have it (don't link to the API code here, send that in Slack if it's in there).

haardikdharma10:
@andyjeffries it is here - https://github.com/civo/civogo/blob/master/instance.go#L169 

Wondering should it be bionic or focal?

andyjeffries:
Should be Focal for now, the latest LTS version of Ubuntu (until Jammy is released - 22.4)

andyjeffries:
Thanks, had forgotten about civogo being in the mix.

alejandrojnm (approved):
LGTM

PR #220, opened 2022-04-13
Draft: CLI commands for PaaS

haardikdharma10:
This MR aims to implement CLI commands for PaaS.


alejandrojnm:
LGTM but you already tested ??

cmd/application_remote.go
```diff
@@ -0,0 +1,86 @@
...
+		// Check if current remote exists.
+		remote_cmd := exec.Command("git", "ls-remote")
+		remote_cmd.Stdout = &stdout
+		remote_cmd.Stderr = &stderr
+		err = remote_cmd.Run()
+		if err != nil {
+			fmt.Printf("error running: git %s\n", strings.Join(remote_cmd.Args, " "))
+			fmt.Println(stderr.String())
+		} else {
+			//What should be the condition here?
```
RealHarshThakur, line 70: Check whether the remote we want to create already exists or not

cmd/application_remote.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/civo/cli/utility"
+	"github.com/spf13/cobra"
+)
+
+var remoteName, remoteURL string
+
+var appRemoteCmd = &cobra.Command{
+	Use:     "remote",
+	Short:   "Add a remote to your current git repository",
+	Example: "civo app remote APP_NAME -r REMOTE_NAME -u REMOTE_URL",
```
RealHarshThakur: I think the URL should be autogenerated.

haardikdharma10: How do we plan to give the URL here then - `remote_cmd = exec.Command("git", "remote", "add", remoteName, remoteURL)`

RealHarshThakur: Remote URL should be : git@git.civo.app:$region/$accountID/$app-name

cmd/application_remote.go
haardikdharma10: How will we do that in code?

cmd/application_remote.go
haardikdharma10: How will we get the Account ID from the API key?

cmd/application.go
```diff
@@ -0,0 +1,75 @@
...
+	Short:   "Configure your application",
+	RunE: func(cmd *cobra.Command, args []string) error {
+		err := cmd.Help()
+		if err != nil {
+			return err
+		}
+		return errors.New("a valid subcommand is required")
+	},
+}
+
```
RealHarshThakur, line 47: Nit: Can you move the scale subcommand to be a part of this file? The inner `Run` function can remain in a separate file but it'll be nice  to stick with the pattern fo initilasing all sub-commands from this file.

cmd/application_config_set.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+
+		findApp, err := client.FindApplication(args[0])
+		if err != nil {
+			utility.Error("%s", err)
+			os.Exit(1)
+		}
+
+		updatedConfig := civogo.EnvVar{
+			Name:  configName,
```
RealHarshThakur, line 41: If configName is "PORT", can you error out with a message: "PORT is an immutable field  picked by Civo"

cmd/application_config_show.go
```diff
@@ -0,0 +1,47 @@
...
+		}
+
+		ow := utility.NewOutputWriter()
+		for _, config := range app.Config {
+			fmt.Println(config)
+		}
+
+		switch outputFormat {
+		case "json":
+			ow.WriteMultipleObjectsJSON(prettySet)
```
RealHarshThakur, line 39: I don't think json output will work if you're doing a `fmt.Println` above. You will have to write data to `ow`

cmd/application_config_unset.go
```diff
@@ -0,0 +1,61 @@
...
+			os.Exit(1)
+		}
+
+		findApp, err := client.FindApplication(args[0])
+		if err != nil {
+			utility.Error("%s", err)
+			os.Exit(1)
+		}
+
+		config := &civogo.UpdateApplicationRequest{
```
RealHarshThakur, line 35: I don't understand how this will unset the env variable. Won't we have to loop through the existing variables, remove the one the user wants and pass that list in the API call?

cmd/application_create.go
```diff
@@ -0,0 +1,127 @@
...
+			s.Prefix = fmt.Sprintf("Creating application (%s)... ", resp.Name)
+			s.Start()
+
+			for stillCreating {
+				application, err = client.FindApplication(resp.Name)
+				if err != nil {
+					utility.Error("%s", err)
+					os.Exit(1)
+				}
+				if application.Status == "ACTIVE" {
```
RealHarshThakur:
```suggestion
				if  strings.ToLower(application.Status) == "available" {
```

cmd/application_domain_add.go
```diff
@@ -0,0 +1,61 @@
...
+			os.Exit(1)
+		}
+
+		findApp, err := client.FindApplication(args[0])
+		if err != nil {
+			utility.Error("App %s", err)
+			os.Exit(1)
+		}
+
+		application := &civogo.UpdateApplicationRequest{
```
RealHarshThakur, line 53: Some validation to see if it's a valid domain/hostname would be nice

cmd/application_domain_remove.go
```diff
@@ -0,0 +1,65 @@
...
+			os.Exit(1)
+		}
+
+		findApp, err := client.FindApplication(args[0])
+		if err != nil {
+			if errors.Is(err, civogo.ZeroMatchesError) {
+				utility.Error("sorry there is no %s application in your account", utility.Red(args[0]))
+				os.Exit(1)
+			}
+			if errors.Is(err, civogo.MultipleMatchesError) {
```
RealHarshThakur: This should be in an `else if` block and in the end, we should have an `else` block for any unanticipated errors

cmd/application_domain_remove.go
```diff
@@ -0,0 +1,65 @@
...
+				utility.Error("sorry we found more than one application with that name in your account")
+				os.Exit(1)
+			}
+		}
+
+		for _, appDomain := range findApp.Domains {
+			if appDomain == args[1] {
+				remove(findApp.Domains, appDomain)
+				utility.Info("Domain %s removed from %s", utility.Green(appDomain), utility.Green(args[0]))
+			}
```
RealHarshThakur, line 74:
```suggestion
break 
			}
```

cmd/application_list.go
```diff
@@ -0,0 +1,57 @@
...
+
+		ow := utility.NewOutputWriter()
+		for _, app := range applications.Items {
+			ow.StartLine()
+			ow.AppendDataWithLabel("id", app.ID, "ID")
+			ow.AppendDataWithLabel("name", app.Name, "Name")
+			ow.AppendDataWithLabel("size", app.Size, "Size")
+			ow.AppendDataWithLabel("network_id", app.NetworkID, "Network ID")
+			ow.AppendDataWithLabel("domains", strings.Join(app.Domains, " "), "Domains")
+			fmt.Println(app.ProcessInfo)
```
RealHarshThakur: Will this then show up when people try to look at output in json?

cmd/application_remove.go
```diff
@@ -0,0 +1,90 @@
...
+			app, err := client.FindApplication(args[0])
+			if err != nil {
+				if errors.Is(err, civogo.ZeroMatchesError) {
+					utility.Error("sorry there is no %s application in your account", utility.Red(args[0]))
+					os.Exit(1)
+				}
+				if errors.Is(err, civogo.MultipleMatchesError) {
+					utility.Error("sorry we found more than one application with that name in your account")
+					os.Exit(1)
+				}
```
RealHarshThakur, line 41: Similar error checking issue as pointed above in domain_remove command

cmd/application_scale.go
```diff
@@ -0,0 +1,62 @@
...
+			os.Exit(1)
+		}
+
+		processInfo := civogo.ProcessInfo{
+			ProcessType:  processType,
+			ProcessCount: processCount,
+		}
+
+		application := &civogo.UpdateApplicationRequest{
+			ProcessInfo: append(findApp.ProcessInfo, processInfo),
```
RealHarshThakur: I don't think appending is the right way to go here. Imagine a user already has a  ProcessType `worker` defined with count 1 but wanted to scale it to 3. Appending would essentially set it to 4(rather than 3).

cmd/application_scale.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/spf13/cobra"
+)
+
+var processType string
+var processCount int
+
+var appScaleCmd = &cobra.Command{
+	Use:     "scale",
+	Aliases: []string{"change", "modify", "upgrade"},
+	Example: "civo app scale APP-NAME --process-type=web --process-count=3",
```
RealHarshThakur: Can we make it possible to do this? 
```suggestion
	Example: "civo app scale APP-NAME web=3 worker=5",
```

cmd/application_scale.go
```diff
@@ -0,0 +1,62 @@
...
+			utility.Error("Creating the connection to Civo's API failed with %s", err)
+			os.Exit(1)
+		}
+
+		findApp, err := client.FindApplication(args[0])
+		if err != nil {
+			utility.Error("App %s", err)
+			os.Exit(1)
+		}
+
```
RealHarshThakur, line 22: Can we validate the process names? Supported types are:
* web
* worker
* release
* clock

We can error out for any other process type user gives.

haardikdharma10: Okay this is done

cmd/application_domain_remove.go
haardikdharma10: This done too

cmd/application_domain_remove.go
haardikdharma10: Done!

cmd/application_list.go
haardikdharma10: My bad, that was me testing something. Editing.

cmd/application_scale.go
haardikdharma10: I don't think. How will we go about so many args? It'll be complex like `arg(process_type)=arg(process_count)`. Also, we support 4 process types as you mentioned which means a series of such args. Is it possible?

cmd/application_config_show.go
haardikdharma10: Done.

cmd/application_scale.go
haardikdharma10: What would you suggest here, then?

cmd/application_remote.go
```diff
@@ -0,0 +1,88 @@
...
+			fmt.Printf("error running: git %s\n", strings.Join(remote_cmd.Args, " "))
+			fmt.Println(stderr.String())
+		} else {
+			//What should be the condition here?
+			fmt.Println("git remote already exists")
+			os.Exit(1)
+		}
+
+		//Add a new remote.
+		remote_cmd = exec.Command("git", "remote", "add", remoteName, fmt.Sprintf("git@git.civo.app:%s/%s/%s", client.Region, client.APIKey, findApp.Name))
```
RealHarshThakur: We cannot give away API key in the git remote URL. This would be a major security flaw 😨 
We need to make a call to `/v2/accounts` , get the accountID

haardikdharma10: Yeah yeah, I know. That is the reason I commented below how to get the Account ID from the API key? 😅

cmd/application_remote.go
RealHarshThakur: Maybe it's fine. We can just error out when they do that.

cmd/application_scale.go
RealHarshThakur: Loop through the existing keys in the array and over ride the value. If the key isn't present, create one

cmd/application_scale.go
RealHarshThakur: You can access all args by doing a `os.Args` . Then you can run a for loop with len(os.Args). Each arg should be in the format of `string=string`, so you can use strings.Split(). After that, you have to validate the process type.

cmd/application_remote.go
haardikdharma10: Okay this is done. have a look at - https://github.com/civo/civogo/pull/90

cmd/application_domain_add.go
RealHarshThakur: Can we validate it this way? It uses a regular expression: https://github.com/asaskevich/govalidator/blob/f21760c49a8d/validator.go#L675

haardikdharma10: Yes, done.

RealHarshThakur: Sorry, I don't see those changes yet.

cmd/application_scale.go
haardikdharma10: When you say each arg will be of format `string=string`, the APP-NAME arg doesn't satisy that?

haardikdharma10: strings.Split will give us the slice like so - e.g [web 3], how do we proceed then? How do we set the process count?

RealHarshThakur: You'll have to first create an empty slice with ProcInfo : `[]ProcInfo`
Once you have the slice[web, 3], you can append it to the ProcInfo slice . 
Something like this:
`procInfo := append(procInfo, ProcessInfo{ name: stringSlice[0], count: stringSlice[1] } `

cmd/application.go
haardikdharma10: This is done in the latest commit.

cmd/application_domain_add.go
haardikdharma10: You can find those in the latest commit now.

cmd/application.go
```diff
@@ -0,0 +1,153 @@
...
+var appScaleCmd = &cobra.Command{
+	Use:     "scale",
+	Aliases: []string{"change", "modify", "upgrade"},
+	Example: "civo app scale APP-NAME PROCESS-NAME=PROCESS-COUNT",
+	Short:   "Scale processes of your application",
+	Args:    cobra.MinimumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
```
RealHarshThakur: Store the function in another file but invoke that function here. 

`Run : scaleCmd()`

cmd/application_config_unset.go
RealHarshThakur: Any update on this?

cmd/application.go
haardikdharma10: Done.

cmd/application_scale.go
```diff
@@ -0,0 +1,79 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/civo/civogo"
+	"github.com/civo/cli/config"
+	"github.com/civo/cli/utility"
+	"github.com/spf13/cobra"
+)
+
+func ScaleCmd(cmd *cobra.Command, args []string) {
```
RealHarshThakur: nit: I don't think this needs to be exported. However, if it does , linter won't be happy as you haven't provided a comment on top of it.

haardikdharma10: Done.

RealHarshThakur:
PR is moved from fork to a branch on civo's remote. Closing this PR 

https://github.com/civo/cli/pull/224/

PR #234, opened 2022-06-30
Reserved ip

RealHarshThakur:
* PR provides a better structure for subcommand
* Adds IP subcommand
* Changes instance public-ip command functionality. Users can now enable/disable public IP on an instance


cmd/ip/ip_assign.go
```diff
@@ -0,0 +1,80 @@
...
+				os.Exit(1)
+			} else if errors.Is(err, civogo.MultipleMatchesError) {
+				utility.Error("sorry we found more than one IP with that value in your account")
+				os.Exit(1)
+			} else {
+				utility.Error("%s", err)
+			}
+		}
+
+		instance, err := client.FindInstance(args[1])
```
haardikdharma10: Wondering how is this working? We always pass instance name/ID with the `instance` flag and not as an argument right? 🤔

cmd/ip/ip.go
```diff
@@ -0,0 +1,36 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+var name, instance string
+
+// IPCmd manages reserved IPs
+var IPCmd = &cobra.Command{
+	Use:     "ip",
+	Aliases: []string{"ips"},
```
haardikdharma10, line 14: We can have more aliases - {ip-address, ip-addresses, reserved-ip, reserved-ips}

cmd/ip/ip_unassign.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/civo/civogo"
+	"github.com/civo/cli/common"
+	"github.com/civo/cli/config"
+	"github.com/civo/cli/utility"
+	"github.com/spf13/cobra"
+)
+
+var ipUnassignCmd = &cobra.Command{
+	Use:     "Unassign",
+	Aliases: []string{"attach"},
```
haardikdharma10: unattach

cmd/ip/ip_assign.go
RealHarshThakur: Fixed

cmd/ip/ip.go
RealHarshThakur: Feels a little counter-intuitive to me that someone would type in the longer name

vishalanarase:
LGTM

vishalanarase (approved):

PR #239, opened 2022-08-10
Fix volume ls command

RealHarshThakur:
Bug: Command failed when it found a volume attached to a cluster worker node as it searched for instance in /v2/instances. 

PR does 3 things:
* Fixed the original bug. 
* Performance improvement by listing resources and looping through them rather than making network calls.
* Update Go version to 1.17

### DIFF
--- a/cmd/volume/volume.go
+++ b/cmd/volume/volume.go
@@ -6,6 +6,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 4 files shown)


alejandrojnm:
can we do this on the API side ???

vishalanarase (approved):
LGTM

alejandrojnm (approved):

PR #240, opened 2022-08-17
Fix k3s Version issue

haardikdharma10:
## Description

Fix k3s version issue to display the stable version when running `civo kubernetes show` command

### DIFF
--- a/cmd/kubernetes_show.go
+++ b/cmd/kubernetes_show.go
@@ -129,20 +129,6 @@ If you wish to use a custom format, the available fields are:
 			ow.WriteKeyValues()
 
 			if kubernetesCluster.UpgradeAvailableTo != "" {
-				var versionsList []string
-				kubeVersions, err := client.ListAvailableKubernetesVersions()
-				if err != nil {
-					utility.Error("%s", err)
-					os.Exit(1)
-				}
-				for _, version := range kubeVersions {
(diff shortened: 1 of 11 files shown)


cmd/kubernetes_show.go
```diff
@@ -139,7 +139,7 @@ If you wish to use a custom format, the available fields are:
 					if version.Type == "deprecated" {
 						continue
 					}
-					if version.Version > kubernetesCluster.Version {
+					if kubernetesCluster.Version != version.Version && version.Type == "stable" {
 						versionsList = append(versionsList, version.Version)
```
RealHarshThakur: I don't think we should restrict them to "stable" versions. There could be development versions available

haardikdharma10: What do you suggest the fix to be then? 🤔

haardikdharma10: Something like `if kubernetesCluster.Version != version.Version && (version.Type == "stable" || version.Type == "development")` ?

RealHarshThakur: I don't think this bit matters: `(version.Type == "stable" || version.Type == "development")` . There are only three types possible: "development", "stable" and "deprecated". We're doing a continue on "deprecated" . So I don't think this is the fix

PR #245, opened 2022-08-20
replacing full stops (.) by colons (:)

saksham31:
made some changes by replacing full stops by colons where required.


DMajrekar:
Thanks for this

Issue #231, opened 2022-06-21
`civo <resourcetype> ls` should show output even if empty


kaihoffman:
Currently, if you have selected a region where you have no resources, you may get an empty response to a listing of resources:

```console
$ civo instance ls
$
```

This is confusing especially if you do not know you have changed regions.

It might be helpful to have a note about which region you're in when listing a region with an empty output. Compare to output from a listing with instances running in it, which shows the region:

```console
$ civo instance ls
+--------------------------------------+--------------+--------+----------+-----------+------+----------+---------------+-------------+--------+
| ID                                   | Hostname     | Region | Size     | Cpu Cores | Ram  | SSD disk | Public IP     | Private IP  | Status |
+--------------------------------------+--------------+--------+----------+-----------+------+----------+---------------+-------------+--------+
| aa068118-40ac-4fc8-b5ea-fb5202a2061a |   hostname   | LON1   | g3.small |         1 | 2048 |       25 | 12.34.56.78 | 192.168.1.2 | ACTIVE |
+--------------------------------------+--------------+--------+----------+-----------+------+----------+---------------+-------------+--------+
```

I suggest having something like

```console
$ civo instance ls
Current region is NYC1. No instances found.
```

kitarp29:
Hi @kaihoffman @andyjeffries @DMajrekar I would love to work on this.
Could you assign this issue to me?

kaihoffman:
@kitarp29 did you have a chance to look at this? Would love to be able to close this off!

kitarp29:
Hi @kaihoffman 
Sorry for the delay got stuck with University exams.
Could you direct me to right file or location I will find the output to the commands?

kaihoffman:
It would need to determine whether the resourcetype is found, and determine the output accordingly @kitarp29 .

For example, the `instance ls` command output is at https://github.com/civo/cli/blob/master/cmd/instance/instance_list.go but it would need to apply equally to other resource types that can be listed.

kitarp29:
If the fix I applied here works for you, I will replicate for the other commands as well :)

kitarp29:
Hey @kaihoffman Could have a look into this
It's not a big breaking change, if you could conform we can get this done sooner.
- https://github.com/civo/cli/pull/246#issuecomment-1221400058

kaihoffman:
@kitarp29 I'm not the best placed to judge a golang PR in our team, I'm afraid - I'll leave it to @alejandrojnm or @RealHarshThakur to have a look?

[...]

Issue #233, opened 2022-06-27
Logging Setup


kitarp29:
I was looking towards to contributing to the project.
I can't seem to find any kind of Logging setup in the project.
Not a pro in Golang but surely have worked around ```log``` module and ```logrus``` module.
I know the benefits of logs. It helps a lot while debugging.
If this is something viable to be added to the project.
Please assign it to me :)
cc: @andyjeffries @DMajrekar @jaysonsantos

kitarp29:
@RealHarshThakur @alejandrojnm 
If [logrus](https://pkg.go.dev/github.com/sirupsen/logrus) seems like a good fit let me.
It has different categories for Logs and will come in real handy if you have any logging architecture setup with something like Fluentd or ELK stack
A good article on the same [Here](https://betterprogramming.pub/awesome-logging-in-go-with-logrus-70606a49f285)

alejandrojnm:
I'm more than happy to use logrus

kitarp29:
If this is something that is needed. assign me to the issue, I will start working on it during the weekend.

RealHarshThakur:
Thanks for assiting with this @kitarp29  
We'll need to make sure that our outputs are consistent from UX though. Shout out to us if you need help

alejandrojnm:
@kitarp29 it will be nice if we can run in debug mode like `kubectl` that shows all the trace if you pass `-v 10`, in that way we can debug every query made by the CLI

RealHarshThakur:
For the initial PR though, happy to just see it be a drop-in replacement for what we have right now.

kitarp29:
Yes, @RealHarshThakur I will be making smaller commits itself so that it will be easier for the maintainers to review it.

uzaxirr:
@kitarp29  Can i take this forward?

kitarp29:
@uzaxirr I am stuck with some work in life. Surely you can take this forward.
If you have any trouble understanding the approach I took, do let me know.
I will help you in understanding. 
The implementation is pretty straightforward, just replicate the work over other functions.

uzaxirr:
> @uzaxirr I am stuck with some work in life. Surely you can take this forward. If you have any trouble understanding the approach I took, do let me know. I will help you in understanding. The implementation is pretty straightforward, just replicate the work over other functions.

Sure thing!

uzaxirr:
@RealHarshThakur Can you assign this to me.

uzaxirr:
[...]

PR #246, opened 2022-08-20
Issue-231 Fixed no Instances found

kitarp29:
- #231 
Hi @kaihoffman this is an easy and quick fix for the issue.
I wanted to break the loop in the length of the array is zero.
But I am concerned about the ``` common.OutputFormat``` condition .
Please review and let me know of the changes.
Signed-off-by: kitarp29 <kitarpsinghrajpoot@gmail.com>

### DIFF
--- a/cmd/instance/instance_list.go
+++ b/cmd/instance/instance_list.go
@@ -1,6 +1,7 @@
 package instance
 
 import (
+	"fmt"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


alejandrojnm (commented):
Hi @kitarp29 thanks for the PR, but more than this you can return a message before  the loop like in the line https://github.com/civo/cli/blob/cc8f88e101bbfdcf3d60c4be2d8378fa18c1c802/cmd/instance/instance_list.go#L60 in that way we can avoid the execution of the rest of the code, it can be something like this 

```golang
if len(instance) == 0 {
	utility.Info("We don't find any instance in the region %s", client.Region)
	os.Exit(1)
}
```

kitarp29:
Yes @alejandrojnm as I mentioned in my first message I knew that the rest of the code did not make sense to execute if ```len(instance)==0```.
So should I commit the changes you mentioned or just add a ``` break``` at the end of my code snippet?
cc: @kaihoffman

cmd/instance/instance_list.go
```diff
@@ -61,7 +61,9 @@ If you wish to use a custom format, the available fields are:
 		ow := utility.NewOutputWriter()
 		for _, instance := range instances {
 			ow.StartLine()
-
+			if len(instances) == 0 {
+				ow.AppendDataWithLabel("region", client.Region, "No instances found in this region")
```
RealHarshThakur: I don't think it needs to  be a ow.Append. Can you please change this to just fmt.Println statement and mention the region in it?

RealHarshThakur:
@kitarp29  thanks for the PR :)

cmd/instance/instance_list.go
```diff
@@ -61,7 +61,9 @@ If you wish to use a custom format, the available fields are:
 		ow := utility.NewOutputWriter()
 		for _, instance := range instances {
 			ow.StartLine()
-
+			if len(instances) == 0 {
```
RealHarshThakur: I think this should move before we run the for loop. Otherwise, this would never get executed

kitarp29: Fixed this 👍🏼 
Put up the for loop in the else statement for the if

cmd/instance/instance_list.go
kitarp29: I have switched it to fmt.Println
Mentioned the region as well

alejandrojnm:
We need to keep a standard in all the code, we can use `println` in some place and in others use `utility.Info`

RealHarshThakur:
Good point @alejandrojnm  . We should move to a well-maintained logging package at some point like @kitarp29  pointed out in this issue. https://github.com/civo/cli/issues/233

cmd/instance/instance_list.go
```diff
@@ -59,39 +60,41 @@ If you wish to use a custom format, the available fields are:
 		}
 
 		ow := utility.NewOutputWriter()
-		for _, instance := range instances {
-			ow.StartLine()
+		if len(instances) == 0 {
+			fmt.Println("region", client.Region, ".No instances found in this region")
+		} else {
```
RealHarshThakur: We don't need the else block here. If it's not 0, rest of the code is executed either way

cmd/instance/instance_list.go
```diff
@@ -59,39 +60,41 @@ If you wish to use a custom format, the available fields are:
 		}
 
 		ow := utility.NewOutputWriter()
-		for _, instance := range instances {
-			ow.StartLine()
+		if len(instances) == 0 {
+			fmt.Println("region", client.Region, ".No instances found in this region")
```
RealHarshThakur:
```suggestion
			fmt.Println("No instances found in %s", client.Region)
			os.Exit(0)
```

cmd/instance/instance_list.go
kitarp29: Removed the else block 👍🏼

cmd/instance/instance_list.go
kitarp29: Did the changes as directed 👍🏼

RealHarshThakur:
@alejandrojnm  Do we want to hold this off until we resolve the logging issue?

alessandroargentieri:
Closing because stale for more than 2 years and with conflicts.

PR #251, opened 2022-08-26
Setting up Logging

kitarp29:
- #233 
Working on this using this [logrus](https://github.com/sirupsen/logrus) module.
There are multiple options and methods that ```logrus``` provides. 
Selecting the ones that the project needs is a topic of discussion,
### What I have done so far:
- Utilized the ```cmd``` module to create different forms of logs. For example, WARN, ERROR, FATAL.
- The logs will be in JSON Format.
- All the logs will contain information about their calling method (as in the function and the file it triggered from).
### What I need help with:
- Figuring out the main file where all these config options of ```logrus``` will be defined, I really this function should be the function that starts as soon as the project initiates.
- Setting up locally to run and test on my system.
- A small skim of the codebase would have been easier, but since a small codebase. I will make do anyway
### Topics for discussing:
- ```logrus``` can write generate all the logs in stdout, stderr or can be used to make ```.log``` or even ```.txt``` filled.
- Other than **FATAL**, **ERROR**, and **WARN** do we need to have **INFO** logs as well?
- We can add colors to logs as well, is it needed? I see there is a utility module in the project. Unsure of its usage as of now.
- What are the fields we need in the logs?
- Are there any other middleware we plan to include in the project for logging purposes?

cc: @alejandrojnm @RealHarshThakur @kaihoffman


kitarp29:
@RealHarshThakur @alejandrojnm ANy updates or take on this?
I will start to apply the changes to other files as well after your green flag.

alejandrojnm:
Can you show the out put ???

kitarp29:
Hi, @alejandrojnm I don't have my laptop right now.
It won't be much different, to be honest. It will be a simple JSON, I will share as I get time.

haardikdharma10:
Closing due to inactivity.

PR #253, opened 2022-08-30
Add CLI commands for object store and object store creds

haardikdharma10:
## Description

This PR adds CLI commands for Object Store and Object Store Credentials.


alejandrojnm:
The build fail, can you check this @haardikdharma10

Issue #232, opened 2022-06-21
Use a defined environment variable in place of `apikey` if it exists


kaihoffman:
Instead of requiring `civo apikey save` for every environment, have it so that the system will use the API key defined in an environment variable if it exists. This means you would be able to configure the env var for a CI/CD system once, download Civo CLI and have it automatically look at the defined variable for the appropriate key to use.

Suggested names for the variable:

- CIVO_APIKEY
- CIVO_TOKEN

(proposed by @rawkode)

alejandrojnm:
This is done

kaihoffman:
Which of the variable names does it look for @alejandrojnm ? Just so we can update the README

alejandrojnm:
So sorry @kaihoffman I forget to mention we go with CIVO_TOKEN

kaihoffman:
Has this been built into a release @alejandrojnm @satakshigarg ? I don't see the documentation on the README and the latest release seems to have been made 23 days ago.

alejandrojnm:
I think this is will be in the next release. We need update the doc.

kaihoffman:
I could update the docs but with the merge having happened it'd be weird being in the readme if that's not how the app works.

alejandrojnm:
You can add and create a PR and add to master, and when the release come out it will be there

kaihoffman:
That's what i mean though, will it be merged at the same time as you cut the release? Because if it's before, I might as well push it now?

PR #254, opened 2022-08-30
Allow user to use defined env key

satakshigarg:
Updated config file to use a defined environment variable in place of API key if it exists.
https://github.com/civo/cli/issues/232


alejandrojnm (approved):

alejandrojnm:
Close #232

Issue #249, opened 2022-08-25
Creating a firewall rule requires "action" despite help text saying there is a default


kaihoffman:
In the `civo firewall rule create` help, there is a note that `--action` defaults to "allow":

`  -a, --action string      the action of the rule can be allow or deny (default is allow) (default "allow")`

However if you run a command without `--action` it fails with

`Error: required flag(s) "action" not set`

Either the text about the default value must be removed, or the CLI should honour it.

alejandrojnm:
Yes, by default is `allow` but if you set `--action` cobra assumes you pass the flag and in this case, is empty because the flag doesn't have any value. The default is in the case you don't pass the flag

kaihoffman:
That's what I mean, I wasn't including `--action` and it would not work.

```
$ civo firewall rule create cli-demo --startport=22 --direction=ingress -l 'SSH access'
Error: required flag(s) "action" not set
```

version 1.0.32

alejandrojnm:
We need fix this then @satakshigarg can you check this ??

satakshigarg:
Sure @alejandrojnm I'll handle this issue

PR #255, opened 2022-08-30
Firewall Rule Create-Fixed default flags issue

satakshigarg:
Fixed the issue with Firewall Rule Create command to set actions and direction flags as optional.
https://github.com/civo/cli/issues/249


alejandrojnm (approved):
LGTM

PR #267, opened 2022-09-14
Updated Readme file

SaptarshiSarkar12:
Added `thanks to all contributors` section in the Readme file.


alessandroargentieri:
Nice idea.. if the number gets huge we would remove it, but for the moment is more than sustainable

SaptarshiSarkar12:
> Nice idea.. if the number gets huge we would remove it, but for the moment is more than sustainable

OK @alessandroargentieri

PR #269, opened 2022-09-15
Add support to list dangling volumes and warning when user tries to delete cluster with volumes attached.

haardikdharma10:
## Description

This PR adds support to list dangling volumes and displays a warning when user tries to delete cluster with volumes attached.


RealHarshThakur:
Can we also mention on `volume ls` call what a dangling volume means?

cmd/volume/volume.go
```diff
@@ -28,6 +28,8 @@ func init() {
 	VolumeCmd.AddCommand(volumeAttachCmd)
 	VolumeCmd.AddCommand(volumeDetachCmd)
 
+	volumeListCmd.Flags().BoolVarP(&dangling, "dangling", "", false, "List only dangling volumes")
```
RealHarshThakur: Can we add a shortcut flag as `d` for this?

RealHarshThakur (approved):

PR #274, opened 2022-09-26
Add message for users to redownload their kubeconfig

haardikdharma10:
## Description

This PR adds functionality to prompt user to redownload their kubeconfig if their cluster is more than 1 year old.

### DIFF
--- a/cmd/apikey/apikey_save.go
+++ b/cmd/apikey/apikey_save.go
@@ -7,7 +7,6 @@ import (
 	"runtime"
 	"strings"
 	"syscall"
-	"time"
 
 	"github.com/civo/civogo"
 	"github.com/civo/cli/common"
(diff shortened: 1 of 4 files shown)


cmd/apikey/apikey_save.go
```diff
@@ -132,6 +133,27 @@ var apikeySaveCmd = &cobra.Command{
 			config.SaveConfig()
 		}
 
+		if config.Current.Clusters == nil {
```
RealHarshThakur: This logic should be in the config.go file while CLI is reading/loading the config

cmd/apikey/apikey_save.go
```diff
@@ -132,6 +133,27 @@ var apikeySaveCmd = &cobra.Command{
...
+			}
+			clusters, err := client.ListKubernetesClusters()
+			if err != nil {
+				fmt.Println(err.Error())
+				os.Exit(1)
+			}
+			config.Current.Clusters = make(map[string]bool)
+			for _, cluster := range clusters.Items {
+				config.Current.Clusters[cluster.ID] = true
+				if time.Since(cluster.BuiltAt) > (365 * 24 * time.Hour) {
```
RealHarshThakur: I think there's a time.Year, if not you can use time.Day *365 but more importantly, we have to check if it's divisible by the year in which cluster was built.

alessandroargentieri:
Cleaning up the open PRs and issues. Despite being an excellent idea, I'm closing it because stale, unfinished and with conflicts. Can be reopened if needed.

PR #288, opened 2022-11-24
fix: deprecated version of actions - go workflow

Pradumnasaraf:
### Problem 

Both `actions/setup-go@v1` and `actions/checkout@v2` action currently uses Node v12 to run, which is soon going to be deprecated. Here is the GitHub Blog about that [GitHub Actions: All Actions will begin running on Node16 instead of Node12](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/)

### Proposed changes.

Updated both go and checkout actions to the latest version.

### Warning Screenshot

Screenshot 2022-11-24 at 1 30 47 PM

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -8,13 +8,13 @@ jobs:
     steps:
 
     - name: Set up Go
-      uses: actions/setup-go@v1
+      uses: actions/setup-go@v3
       with:
         go-version: 1.18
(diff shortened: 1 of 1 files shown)


Pradumnasaraf:
Hey, @satakshigarg can you have a look?

Thank you

PR #291, opened 2022-11-30
Resolved Github Rate Limit Errors

satakshigarg:
Added error check for Github Rate Limit as well as API response errors

### DIFF
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -40,8 +40,6 @@ var (
 			releases, err := m.LatestReleases()
 			if err != nil {
 				if common.IsGHError(err) != nil {
-					options := "1. Wait a few minutes and try again\n 2. Try switching to a different network (e.g. mobile data, VPN, etc) as GitHub could be rate-limiting your IP address \n 3. Download latest CLI from https://github.com/civo/cli/releases"
-					fmt.Println(options)
 					os.Exit(1)
 				}
 				utility.Error("error fetching releases: %s", err)
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -39,9 +39,9 @@ var (
 			// fetch the new releases
 			releases, err := m.LatestReleases()
 			if err != nil {
-				if common.IsGHError(err) {
+				if common.IsGHError(err) || common.IsGHRatelimitError(err) {
 					options := "1. Wait a few minutes and try again\n 2. Try switching to a different network (e.g. mobile data, VPN, etc)\n 3. Download latest CLI from https://github.com/civo/cli/releases"
-					fmt.Println("You have reached the github rate limit:\n " + options)
+					fmt.Println(options)
 					os.Exit(1)
 				}
 				utility.Error("error fetching releases: %s", err)
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -39,7 +39,7 @@ var (
 			// fetch the new releases
 			releases, err := m.LatestReleases()
 			if err != nil {
-				if common.IsGHError(err) {
+				if common.IsGHError(err) != nil {
 					options := "1. Wait a few minutes and try again\n 2. Try switching to a different network (e.g. mobile data, VPN, etc) as GitHub could be rate-limiting your IP address \n 3. Download latest CLI from https://github.com/civo/cli/releases"
 					fmt.Println(options)
 					os.Exit(1)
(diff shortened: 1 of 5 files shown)


cmd/update.go
```diff
@@ -39,9 +39,9 @@ var (
 			// fetch the new releases
 			releases, err := m.LatestReleases()
 			if err != nil {
-				if common.IsGHError(err) {
+				if common.IsGHError(err) || common.IsGHRatelimitError(err) {
 					options := "1. Wait a few minutes and try again\n 2. Try switching to a different network (e.g. mobile data, VPN, etc)\n 3. Download latest CLI from https://github.com/civo/cli/releases"
```
RealHarshThakur:
```suggestion
					options := "1. Wait a few minutes and try again\n 2. Try switching to a different network (e.g. mobile data, VPN, etc) as GitHub could be rate-limiting your IP address \n 3. Download latest CLI from https://github.com/civo/cli/releases"
```

common/common.go
```diff
@@ -50,3 +50,9 @@ func IsGHError(err error) bool {
 	_, ok := err.(*github.ErrorResponse)
```
RealHarshThakur: Comment on this method:
```suggestion
// IsGHError checks  if any errror from github is returned
```

common/common.go
```diff
@@ -50,3 +50,9 @@ func IsGHError(err error) bool {
...
 	return ok
 }
+
+// IsGHRateLimitError checks if the error is a github rate limit error
+func IsGHRatelimitError(err error) bool {
```
RealHarshThakur: I don't think we need this as we're handling it when there is any error from github

common/common.go
```diff
@@ -45,8 +45,10 @@ func VersionCheck() (res *latest.CheckResponse, skip bool) {
...
 
-// IsGHError checks if the error is a github rate limit error
+// IsGHError checks  if any errror from github is returned
 func IsGHError(err error) bool {
-	_, ok := err.(*github.ErrorResponse)
-	return ok
+	_, rateLimit := err.(*github.RateLimitError)
+	_, abuseRateLimit := err.(*github.AbuseRateLimitError)
+	_, twoFactor := err.(*github.TwoFactorAuthError)
+	return rateLimit || abuseRateLimit || twoFactor
```
RealHarshThakur, line 53: There should be a better way to check if an error is a Github Error.

common/common.go
```diff
@@ -45,8 +45,10 @@ func VersionCheck() (res *latest.CheckResponse, skip bool) {
 	return res, false
 }
 
-// IsGHError checks if the error is a github rate limit error
+// IsGHError checks  if any errror from github is returned
 func IsGHError(err error) bool {
-	_, ok := err.(*github.ErrorResponse)
-	return ok
+	_, rateLimit := err.(*github.RateLimitError)
```
RealHarshThakur: Does the test pass without the changes you made here?

satakshigarg: Yes the tests passed for these errors. But I have made few changes to generalise it to all Github API errors.

cmd/update.go
```diff
@@ -39,9 +39,9 @@ var (
 			// fetch the new releases
 			releases, err := m.LatestReleases()
 			if err != nil {
-				if common.IsGHError(err) {
-					options := "1. Wait a few minutes and try again\n 2. Try switching to a different network (e.g. mobile data, VPN, etc)\n 3. Download latest CLI from https://github.com/civo/cli/releases"
-					fmt.Println("You have reached the github rate limit:\n " + options)
+				if common.IsGHError(err) != nil {
+					options := "1. Wait a few minutes and try again\n 2. Try switching to a different network (e.g. mobile data, VPN, etc) as GitHub could be rate-limiting your IP address \n 3. Download latest CLI from https://github.com/civo/cli/releases"
```
RealHarshThakur: Should we be printing the error?

common/common.go
```diff
@@ -36,7 +38,7 @@ func VersionCheck() (res *latest.CheckResponse, skip bool) {
 	}
 	res, err := latest.Check(githubTag, strings.Replace(VersionCli, "v", "", 1))
 	if err != nil {
-		if IsGHError(err) {
+		if IsGHError(err) != nil {
```
RealHarshThakur, line 41: Good call, I agree we should not be printing the error in this path.

RealHarshThakur (approved):

PR #293, opened 2023-01-04
Fixed error message for deleting k8 app

satakshigarg:
Replaced the older error message for missing uninstall scripts with better error.

### DIFF
--- a/cmd/kubernetes/kubernetes_app_remove.go
+++ b/cmd/kubernetes/kubernetes_app_remove.go
@@ -70,8 +70,7 @@ var kubernetesAppRemoveCmd = &cobra.Command{
 					output, _ := cmd.CombinedOutput()
 					fmt.Println("--------------- Uninstall script ---------------")
 					fmt.Println(string(output))
-					fmt.Println("--------------- Uninstall output ---------------")
-					fmt.Println(b.String())
+					fmt.Printf("Uninstall script for application %s not found", appName)
 					os.Exit(1)
 				} else {
 					utility.Error("Failed to uninstall application %s because of %s\n", appName, err.Error())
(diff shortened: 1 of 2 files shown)


cmd/kubernetes/kubernetes_app_remove.go
```diff
@@ -70,8 +70,7 @@ var kubernetesAppRemoveCmd = &cobra.Command{
 					output, _ := cmd.CombinedOutput()
 					fmt.Println("--------------- Uninstall script ---------------")
 					fmt.Println(string(output))
-					fmt.Println("--------------- Uninstall output ---------------")
-					fmt.Println(b.String())
+					fmt.Printf("Uninstall script for application %s not found", appName)
```
alessandroargentieri: This returns always the "not found" human-readable message even if the uninstall is executed correctly.
We expect, instead, something like:
```golang
fmt.Println("--------------- Uninstall output ---------------")
if strings.Contains(b.String(), "command not found") {
     fmt.Printf("Uninstall script for application %s not found\n", appName)
} else {
    fmt.Println(b.String())
}
```
When using `fmt.Printf` remember to add '\n' at the end of the line otherwise the logs are chained on the same line

alessandroargentieri (changes requested):
Just a change

cmd/kubernetes/kubernetes_app_remove.go
satakshigarg: @alessandroargentieri sure, I have added the suggestions and will make sure to add newline with printf in future.

satakshigarg:
> Just a change

Fixed

alessandroargentieri (approved):
LGTM

PR #294, opened 2023-01-09
Renamed version flag for k8s create

satakshigarg:
Version flag doesn't appear` civo kubernetes create --help` command.

### DIFF
--- a/cmd/kubernetes/kubernetes.go
+++ b/cmd/kubernetes/kubernetes.go
@@ -78,7 +78,7 @@ func init() {
 	kubernetesCreateCmd.Flags().StringVarP(&targetNodesSize, "size", "s", "g4s.kube.medium", "the size of nodes to create.")
 	kubernetesCreateCmd.Flags().StringVarP(&networkID, "network", "t", "default", "the name of the network to use in the creation")
 	kubernetesCreateCmd.Flags().IntVarP(&numTargetNodes, "nodes", "n", 3, "the number of nodes to create (the master also acts as a node).")
-	kubernetesCreateCmd.Flags().StringVarP(&kubernetesVersion, "version", "v", "latest", "the k3s version to use on the cluster. Defaults to the latest. Example - 'civo k3s create --version 1.21.2+k3s1'")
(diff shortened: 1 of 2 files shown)


cmd/kubernetes/kubernetes.go
```diff
@@ -78,7 +78,7 @@ func init() {
 	kubernetesCreateCmd.Flags().StringVarP(&targetNodesSize, "size", "s", "g4s.kube.medium", "the size of nodes to create.")
 	kubernetesCreateCmd.Flags().StringVarP(&networkID, "network", "t", "default", "the name of the network to use in the creation")
 	kubernetesCreateCmd.Flags().IntVarP(&numTargetNodes, "nodes", "n", 3, "the number of nodes to create (the master also acts as a node).")
-	kubernetesCreateCmd.Flags().StringVarP(&kubernetesVersion, "version", "v", "latest", "the k3s version to use on the cluster. Defaults to the latest. Example - 'civo k3s create --version 1.21.2+k3s1'")
+	kubernetesCreateCmd.Flags().StringVarP(&kubernetesVersion, "k3s-version", "v", "latest", "the k3s version to use on the cluster. Defaults to the latest. Example - 'civo k3s create --k3s-version 1.21.2+k3s1'")
```
alessandroargentieri: I would keep "version" by specifying the missing clustername in the example like this here:
```golang
kubernetesCreateCmd.Flags().StringVarP(&kubernetesVersion, "version", "v", "latest", "the k3s version to use on the cluster. Defaults to the latest. Example - 'civo k3s create mycluster --version 1.21.2+k3s1'")
```
and transform the `version` global `PersistentFlag` into a normal flag (this causes the problem):
(in root.go):
```golang
rootCmd.Flags().BoolVarP(&version, "version", "v", false, "Print the version of the CLI")
```

alessandroargentieri (changes requested):
The version command in `civo kubernetes create` is missing in the docs because the `civo --version` flag is wrongly declared as a persistent flag, which means a flag that has the same behaviour at every level.

cmd/kubernetes/kubernetes.go
satakshigarg: @alessandroargentieri  if we convert --version  to non-global flag, it won't print in any of the documentations other than `civo kubernetes create --help`. I think we should not convert it to global(and remove it from all other commands) just for making it appear for one command.

alessandroargentieri (approved):
LGTM

PR #298, opened 2023-01-31
Fix error in the CLI and the GitHub API

alejandrojnm:
We fix the error in the CLI to prevent making multiple calls to Github API to get the last version, now we check the version every 24h.


haardikdharma10 (approved):
LGTM

Issue #207, opened 2022-03-14
Use cli to show current cost 


stiliajohny:
hi 

Would it be a good idea to have a flag for the `civo` cli to show the current cost of the account the API key has access to ? 
Potentially add the permission as well as part of the implementation

stiliajohny:
+1 on this

Arijit1000:
Hello @stiliajohny this is actually a good idea to show the current cost of the account in the `civo` cli itself

stiliajohny:
I wonder if it is something that I going to be implemented

uzaxirr:
@RealHarshThakur 
Is this possible? does there exists any Endpoint which shows the cost

RealHarshThakur:
Not yet but we plan to have an API endpoint for it. Can't give an ETA on it though.

uzaxirr:
> Not yet but we plan to have an API endpoint for it. Can't give an ETA on it though.

So this feature can not be implemented until that API comes into play right?

PR #309, opened 2023-04-07
cost-cli added civogo-GetCost pending

Horiodino:
https://github.com/civo/cli/issues/207

### DIFF
--- a/cmd/cost/cost.go
+++ b/cmd/cost/cost.go
@@ -1,0 +1,54 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/civo/civogo"
+	"github.com/civo/cli/config"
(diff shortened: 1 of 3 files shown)


cmd/cost/cost.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+
+		// If there is an error, display it and exit
+		if err != nil {
+			fmt.Println(err)
+			os.Exit(1)
+		}
+
+		// client is already defined, so we can just call the GetCost() method
+		cost, err := client.GetCost()
```
uzaxirr, line 38: Does this function even exists in the SDK??
I couldn’t find it in the docs. and ive also checked https://www.civo.com/api i couldn’t find any endpoint which satisfies the current requirements.

Issue #303, opened 2023-03-02
Slow firewall list command


zulh-civo:
When I sent `curl` to list all firewalls ([docs](https://www.civo.com/api/firewall#list-firewalls)), the result was fast. But when I did `civo firewall ls`, it's very slow. I'm not entirely sure what's causing this and whether this affect others too or just me. Can someone from CLI team check?

**Version**

```
$ civo version
Civo CLI v1.0.48
```

**STAGING**

> 1.9 seconds vs 1 minute 55 seconds

```
$ time curl --silent --location --request GET 'https://<staging-host>/v2/firewalls?region=STAGING' \
--header 'Authorization: Bearer <redacted>'
...
curl --silent --location --request GET  --header   0.02s user 0.01s system 1% cpu 1.964 total

$ time civo firewall ls
...
civo firewall ls  0.34s user 0.22s system 0% cpu 1:55.26 total
```

**LON1**

> 4.4 seconds vs 5 minutes 47 seconds

```
$ time curl --silent --location --request GET 'https://<production-host>/v2/firewalls?region=LON1' \
--header 'Authorization: Bearer <redacted>'
...
curl --silent --location --request GET  --header   0.03s user 0.02s system 0% cpu 4.400 total

$ time civo firewall ls
...
civo firewall ls  0.38s user 0.24s system 0% cpu 5:47.76 total
```

RealHarshThakur:
This is a CLI issue. API latencies are within our SLI

haardikdharma10:
This should be fixed with the above PR^ 
Closing.

PR #318, opened 2023-05-17
Fix slow firewall list command

haardikdharma10:
## Background

https://github.com/civo/cli/issues/303

### DIFF
--- a/cmd/firewall/firewall_list.go
+++ b/cmd/firewall/firewall_list.go
@@ -1,6 +1,7 @@
 package firewall
 
 import (
+	"github.com/civo/civogo"
 	"github.com/civo/cli/common"
 	"github.com/civo/cli/config"
 	"github.com/civo/cli/utility"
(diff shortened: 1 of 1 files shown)


cmd/firewall/firewall_list.go
```diff
@@ -1,6 +1,7 @@
 package firewall
 
 import (
+	"github.com/civo/civogo"
```
RealHarshThakur: Import orders are wrong

cmd/firewall/firewall_list.go
```diff
@@ -43,9 +44,20 @@ Example: civo firewall ls -o custom -f "ID: Name"`,
...
+		if err != nil {
+			utility.Error("%s", err)
+			os.Exit(1)
+		}
+
+		var network civogo.Network
 		ow := utility.NewOutputWriter()
 		for _, firewall := range firewalls {
-			network, _ := client.FindNetwork(firewall.NetworkID)
+			for _, network = range networks {
```
RealHarshThakur: I would suggest before looping over firewalls , loop over Networks to create a "map[networkID]:networkLabel" .

When you run the firewall loop, you can directly use it that map to give out network name on L66 rn

RealHarshThakur (approved):

PR #325, opened 2023-07-06
Add check for kfaas availability in a region

haardikdharma10:
## Description

This PR adds check to the CLI for `kfaas` feature availability in a region

Corresponding `civogo` PR - https://github.com/civo/civogo/pull/157

### DIFF
--- a/cmd/kfcluster/kfcluster_create.go
+++ b/cmd/kfcluster/kfcluster_create.go
@@ -21,6 +21,18 @@ var kfcCreateCmd = &cobra.Command{
 	Args:    cobra.MinimumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		utility.EnsureCurrentRegion()
+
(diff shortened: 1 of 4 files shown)


cmd/kfcluster/kfcluster_create.go
```diff
@@ -73,7 +85,6 @@ var kfcCreateCmd = &cobra.Command{
 			NetworkID:  network.ID,
 			Size:       size,
 			FirewallID: firewallID,
-			Region:     client.Region,
```
RealHarshThakur, line 76: why are we removing it?

haardikdharma10: We pass it before sending a create request - https://github.com/civo/civogo/blob/master/kfcluster.go#L113
Same way we do for k3s clusters. No need to add it here.

RealHarshThakur (approved):

PR #343, opened 2023-09-19
fix for create pool

alessandroargentieri:
Create pool fails because still using the PUT `/v2/kubernetes/clusters/{id}` endpoint.
This MR implements the client for the new endpoint `POST /v2/kubernetes/clusters/{id}/pools` for creating new pools.

Already tested in production on the real case which was failing.


alejandrojnm (approved):
LGTM we need to do the same in terraform, I can do it

PR #366, opened 2024-02-23
Add Support for DB version list

uzaxirr:
Screenshot 2024-02-24 at 2 17 49 AM

### DIFF
--- a/cmd/database/database.go
+++ b/cmd/database/database.go
@@ -36,6 +36,7 @@ func init() {
 	DBCmd.AddCommand(dbEngineCmd)
 	DBCmd.AddCommand(dbBackupCmd)
 	DBCmd.AddCommand(dbRestoreCmd)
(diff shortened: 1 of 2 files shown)


alejandrojnm:
Nice, Why did you remove the utility `EnsureCurrentRegion` ?

uzaxirr:
> Nice, Why did you remove the utility `EnsureCurrentRegion` ?

That's bcz db versions are independent of regions and irrespective of the region the response from API will always be the same.

alejandrojnm:
The API reads the version from the backend (region), so that means we can have different versions per region, any way you can confirm this with @alessandroargentieri

alessandroargentieri:
Yes, it depends from the region. So the region must be passed to the api call like we do for any civo resource

uzaxirr:
Done, made the required changes.

alessandroargentieri (approved):
LGTM

PR #369, opened 2024-02-25
Add Support for objectstore_credentials_show.go

uzaxirr:
Screenshot 2024-02-26 at 4 18 35 AM

### DIFF
--- a/cmd/objectstore/objectstore.go
+++ b/cmd/objectstore/objectstore.go
@@ -65,6 +65,7 @@ func init() {
 	objectStoreCredentialCmd.AddCommand(objectStoreCredentialCreateCmd)
 	objectStoreCredentialCmd.AddCommand(objectStoreCredentialUpdateCmd)
 	objectStoreCredentialCmd.AddCommand(objectStoreCredentialDeleteCmd)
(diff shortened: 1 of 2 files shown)


alejandrojnm:
If I pass an invalid name or ID, return not found, a nice message @alessandroargentieri Do you agree ??

uzaxirr:
> If I pass an invalid name or ID, return not found, a nice message @alessandroargentieri Do you agree ??

something like this?
Screenshot 2024-02-26 at 6 43 46 PM

alessandroargentieri:
@kaihoffman as product owner I would like to ask you if we want this behaviour to be added to the CLI. In fact, we have the equivalent:
```bash
$ civo objectstore credential secret --access-key <ACCESS-KEY-ID>
Your secret key is: <SECRET-ACCESS-KEY>
```
By adding the `show` command we are making the `secret` command useless. So maybe removable if we want to move in this direction and change things.

giornetta:
Hey - what's the status on this?

kaihoffman:
Was the `show` command ever added?

giornetta:
@kaihoffman I don't think it was. Do we want to add it and merge this?

uzaxirr:
nope this should closed for now

PR #377, opened 2024-03-29
Add VLAN Support in network create

uzaxirr:
Screenshot 2024-03-30 at 12 26 08 AM


vishalanarase (changes requested):
In the get and list we need to show info if vlan is enabled

uzaxirr:
> In the get and list we need to show info if vlan is enabled

That will be done in a separate PR, #378

PR #378, opened 2024-04-02
Cmd to Describe network

uzaxirr:
Screenshot 2024-04-02 at 11 41 08 PM


alessandroargentieri:
Maybe a format like the one for kubernetes would be better since the number of parameters to be shown is high and variable (gateway_id, vlan_id, harware_addr are missing)

alejandrojnm:
@alessandroargentieri or @uzaxirr  we have `(gateway_id, vlan_id, harware_addr)` in the network ??

alessandroargentieri:
Yes we have. All these params are optional

uzaxirr:
> Maybe a format like the one for kubernetes would be better since the number of parameters to be shown is high and variable (gateway_id, vlan_id, harware_addr are missing)

`gateway_id`, `vlan_id`, `harware_addr` are visible when vlan is available for the given network.

Changed the format to the one being shown below.
Screenshot 2024-04-03 at 5 32 15 AM

alessandroargentieri (dismissed):
LGTM

alejandrojnm:
The new format is different from the rest of the CLI, we wanna that ??

uzaxirr:
> The new format is different from the rest of the CLI, we wanna that ??

nope its not different, objectstore show and kubernetes show also follow the same format

PR #383, opened 2024-04-18
Validate CIDR on firewall_rule_create.go

### LINKED ISSUE #382: Firewall Rule Create Value Checking
Problem description:

The CLI accepts invalid addresses for the --cidr flag.

Example: --cidr=%CIDR

Complete Command:

civo firewall rule create %NAME% ^
--protocol=TCP ^
--startport=1 ^
--endport=65535 ^
--direction=ingress ^
--label="Allow all TCP traffic from home" ^
--action="allow" ^
--cidr=%CIDR

The error is caused by a typo "%CIDR" should be "%CIDR%. In the script file I use environment variable substitution because I create multiple rules in one batch file.

I realize that validating IP addresses can be a pain, but the characters "%CIDR" are not valid for IPv4 addresses. Perhaps the "C" if you accept hex addressing, but I did not test that.

Problem level: nuisance - not critical

Environment:
CLI version: 1.0.80
OS: Windows 10 x64
Command Prompt (not PowerShell)

uzaxirr:
Fixes: #382 
Screenshot 2024-04-18 at 2 30 00 PM

### DIFF
--- a/cmd/firewall/firewall_rule_create.go
+++ b/cmd/firewall/firewall_rule_create.go
@@ -2,6 +2,7 @@ package firewall
 
 import (
 	"fmt"
+	"net"
 	"strings"
 
 	"github.com/civo/civogo"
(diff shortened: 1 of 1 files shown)


alessandroargentieri (approved):
LGTM

PR #408, opened 2024-04-27
Enable Deletion of current API key

### LINKED ISSUE #403: Is not being able to remove the current API key a desired behaviour?
## Issue

You cannot delete the current API key

```bash
% civo apikey remove civo-admin
Warning: The API key "civo-admin" is the current one, please change it before removing it
```

But why? What if it's the only api key and you want to delete it for any reason? I know this can be done by deleting `~/.civo.json` but the user has no way of knowing this, and I can totally see scenarios where you would want to delete all api keys.

## Acceptance Criteria

* Allow the user to delete current API key, or at least provide a flag that lets you do it.

uzaxirr:
Fixes: #403 

Screenshot 2024-04-27 at 11 36 29 AM


alejandrojnm (commented):
I think now have to much confirmation message, @fernando-villalba ??

PR #416, opened 2024-05-08
Honour Local path flag while saving kube configs

uzaxirr:


saiyam1814 (dismissed):

cmd/kubernetes/kubernetes_config.go
```diff
@@ -27,6 +27,7 @@ Notes:
 
 var saveConfig, switchConfig, overwriteConfig bool
 var localPathConfig string
+var defaultKubeConfigPath = "/.kube/config"
```
alessandroargentieri: I think it's wrong. This is an absolute path. Try with `$HOME/.kube/config` but I'm not sure it's going to work in windows

saiyam1814: Good point Ale

alessandroargentieri: We can try something like:
```
var defaultConfigMapPath := "$HOME/.kube/config"
if runtime.GOOS == "windows" {
    defaultConfigMapPath = "%USERPROFILE%\.kube\config"
}
```

alessandroargentieri: or something like:
```
dirname, err := os.UserHomeDir()
 if err != nil {
       log.Fatal( err )
}
defaultConfigMapPath := fmt.Sprintf("%s/.kube/config", dirname)
```

uzaxirr: https://github.com/civo/cli/pull/416/commits/d91755ee4e453add89bd5feb19c03491264f2644

alessandroargentieri (approved):
LGTM

alessandroargentieri (approved):
LGTM

alessandroargentieri:
@uzaxirr I decided to merge and not to use the `os.UserHomeDir()` which would have involved managing an error

Issue #428, opened 2024-05-28
Creating networks with the CLI does not create a default firewall rule for it


fernando-villalba:
## Issue

The default network has a default firewall that cannot be deleted.

When creating another network in the UI, it also creates a default firewall that also cannot be deleted. 

Creating a network in the CLI however does not create this default firewall.

This issue also affects the [Terraform provider](https://github.com/civo/terraform-provider-civo/issues/212).

## Acceptance criteria

* Ensure the behaviour of the CLI matches that of the UI. When a network is created, a default firewall should also be created.

haardikdharma10:
Closing as completed in https://github.com/civo/cli/pull/431

PR #431, opened 2024-06-13
Add default firewall flag for network create command

haardikdharma10:
## Description

- Adds a new flag `--create-default-firewall` for `network create` command to allow users create a default firewall for their networks.

## Background

https://github.com/civo/cli/issues/428


uzaxirr (approved):
LGTM!

Issue #433, opened 2024-06-26
Repo standard/code quality analysis 


SoulPancake:
I think this is a very early to comment on this but perhaps there's a lot of work to be done to make this repo better.

I was only going through the code, one of the very first things that comes as an issue is 
1. The incredibly large readme.md
It's overwhelming to have all that info altogether at one place, instead it should be modularized to point to a bunch of separate docs describing the installation steps or other things, maybe links in some cases.
It probably seems useful to have all the info at one place but it is not dev-friendly at all.

2. Linters please? 
I don't mean to be a snob but without a linter eventually the debt of not maintaining code style can catch up, even though this is possibly only maintained by few people at this point, avoiding this as early as possible is a life-saver. 
![image](https://github.com/civo/cli/assets/70265851/4abb9e7e-1a6d-4648-a6cb-535b4518fc32)

3. Straight up tests not working
![image](https://github.com/civo/cli/assets/70265851/f79e4507-cdbe-47a4-96e9-dc22e3490d0d)
![image](https://github.com/civo/cli/assets/70265851/9e7bfdff-9f2d-4ee4-a929-736aecc6e829)

Apart from having incredibly low coverage, the repo also has some not-working tests.
Having properly tested code will definitely ensure low tech debt over-time.

I understand a lot of this might be tried/tested code and under maintenance but coming from working in very large repos with a significant number of contributors, hope all these points help out to be a large number of 
`good first issues` for the community.

I'm down to contribute to few of these and also help setting up automated checks for these to maintain quality, in this repo as well as others.

Cheers and good luck!
Long live OpenSource.

giornetta:
Hi @SoulPancake, thanks for bringing this up and sorry for the long delay! You brought up lots of valid points, we'll make sure to work on addressing the issues you've raised 😄

fulviodenza:
Hi @SoulPancake and thanks for raising this. I agree with you 100%.
We could evaluate splitting the three points of your issue in 3 more sub-issues:
- Refactoring readme.md
- Linter
- Tests not working

I think all of them could be tagged as good first issue.

giornetta:
@fulviodenza I believe the third issue has already been solved by #527.

Regarding the huge README file, I propose we should move that to a more structured **Wiki** section under this repository.

PR #434, opened 2024-06-26
fix: output writer test

SoulPancake:
Part of #433

### DIFF
--- a/utility/output_writer_test.go
+++ b/utility/output_writer_test.go
@@ -11,7 +11,7 @@ func TestWriteCustomOutput(t *testing.T) {
 }
 
 // TestExampleFWriteCustomOutput shows how to use the WriteCustomOutput function
-func TestExampleFWriteCustomOutput() {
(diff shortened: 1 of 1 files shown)


utility/output_writer_test.go
```diff
@@ -11,7 +11,7 @@ func TestWriteCustomOutput(t *testing.T) {
 }
 
 // TestExampleFWriteCustomOutput shows how to use the WriteCustomOutput function
-func TestExampleFWriteCustomOutput() {
+func TestExampleFWriteCustomOutput(t *testing.T) {
```
pavelnikolov, line 14: The func name must start with `Example` and not with `Test` if you want to use the output. You don't need the `t *testing.T` argument.

giornetta:
Hi @SoulPancake, thanks for noticing and for picking this up! I missed this PR and ended up merging another one that solved the same problem. I will close this.

SoulPancake:
Cool
Sorry I missed following up on this
Thanks

Issue #406, opened 2024-04-26
We need flags to delete all dependent resources and configurations with deleting a cluster


fernando-villalba:
## Issue
Currently when you create and delete clusters, it leaves too many dependent resources and configuration behind, for example:

1. Firewalls
2. Volumes (especially if creating apps) 
3. Kube config

Deleting all of these by hand or scripting it separately is very tedious

We need flags that delete these resources together with the cluster. 

#### 1. I suggest making the deletion of firewalls and kubeconfig the default, and keep volumes by default but provide option to get rid of them

After all, you can always revert your firewalls and config is only valid for that cluster, so no point in keeping them for the vast majority of cases. 

And then have flags to delete the volumes that may have important data. 

So you could have flags like this:

`civo kubernetes delete --delete-volumes` --> Deletes dependent volumes
`civo kubernetes delete --keep-firewalls` --> keeps dependent firewalls
`civo kubernetes delete --keep-kubeconfig` --> keeps dependent firewalls

#### 2. When deleting clusters let the user know what was left behind

Aside outputting what cluster names and ids were deleted, also output the volumes that were left behind, and as an info message sent to stderr, you can mention the flag to delete dependent volumes next time.

alejandrojnm:
Can be a global one like --delete-dependent-resource that by default can be false and if you pass the api can delete all related object, we can use the same flag in terraform

uzaxirr:
@alejandrojnm 
There is already such flag in API which automatically deletes all the related resources?
If yes then can you please specify which API.

alessandroargentieri:
@uzaxirr no we don't have and even the UI is not deleting them. We had some conversation around it multiple times, but they never translated into a practical action.

harsh082ip:
Hey @fernando-villalba @uzaxirr, I can solve this issue. Could you assign it to me?

harsh082ip:
I've submitted a pull request. Please review and let me know if any changes are required. 
@fernando-villalba  @uzaxirr @alejandrojnm @alejandrojnm

PR #437, opened 2024-07-08
Added flags to delete all dependent resources and configurations with…

harsh082ip:
As referenced in issue #406, this update includes the following enhancements:

Flags for Deletion Operations:
--delete-volumes: Deletes associated volumes when deleting a Kubernetes cluster.
--keep-firewalls: Retains dependent firewalls during deletion.
--keep-kubeconfig: Preserves Kubernetes configurations.

Additional Output and Messaging:
Implemented the second part of the acceptance criteria:
Alongside cluster deletion details, outputs information about remaining volumes.
Provides an informational message on stderr, suggesting the --delete-volumes flag for future use.

Error Handling:
Addressed potential error scenarios to ensure robust operation.
Please review and let me know if any changes are needed.

### DIFF
--- a/cmd/kubernetes/kubernetes.go
+++ b/cmd/kubernetes/kubernetes.go
@@ -142,6 +142,11 @@ func init() {
 	kubernetesNodePoolInstanceListCmd.Flags().StringVarP(&nodePoolID, "node-pool-id", "p", "", "the ID of the node pool.")
 
 	kubernetesNodePoolCmd.AddCommand(kubernetesNodePoolListCmd)
+
+	// Define the flags for the kubernetesRemoveCmd
+	kubernetesRemoveCmd.Flags().BoolVar(&deleteVolumes, "delete-volumes", false, "Delete dependent volumes")
+	kubernetesRemoveCmd.Flags().BoolVar(&keepFirewalls, "keep-firewalls", false, "Keep dependent firewalls")
+	kubernetesRemoveCmd.Flags().BoolVar(&keepKubeconfig, "keep-kubeconfig", false, "Keep kubeconfig")
 }
 
 func getKubernetesList(value string) []string {
--- a/cmd/kubernetes/kubernetes_remove.go
+++ b/cmd/kubernetes/kubernetes_remove.go
@@ -14,7 +14,15 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var kuberneteList []utility.ObjecteList
+var (
+	deleteVolumes  bool // Flag to delete dependent volumes
+	keepFirewalls  bool // Flag to keep dependent firewalls
+	keepKubeconfig bool // Flag to keep kubeconfig
+)
+
+var kuberneteList []utility.ObjecteList // List to store the Kubernetes clusters to be deleted
+
+// Command definition for removing a Kubernetes cluster
 var kubernetesRemoveCmd = &cobra.Command{
 	Use:     "remove",
 	Aliases: []string{"rm", "delete", "destroy"},
@@ -30,6 +38,7 @@ var kubernetesRemoveCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		utility.EnsureCurrentRegion()
 
+		// Create a client to interact with the Civo API
 		client, err := config.CivoAPIClient()
 		if common.RegionSet != "" {
 			client.Region = common.RegionSet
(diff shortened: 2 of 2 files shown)


cmd/kubernetes/kubernetes.go
```diff
@@ -142,6 +142,11 @@ func init() {
 	kubernetesNodePoolInstanceListCmd.Flags().StringVarP(&nodePoolID, "node-pool-id", "p", "", "the ID of the node pool.")
 
 	kubernetesNodePoolCmd.AddCommand(kubernetesNodePoolListCmd)
+
+	// Define the flags for the kubernetesRemoveCmd
+	kubernetesRemoveCmd.Flags().BoolVar(&deleteVolumes, "delete-volumes", false, "Delete dependent volumes")
+	kubernetesRemoveCmd.Flags().BoolVar(&keepFirewalls, "keep-firewalls", false, "Keep dependent firewalls")
+	kubernetesRemoveCmd.Flags().BoolVar(&keepKubeconfig, "keep-kubeconfig", false, "Keep kubeconfig")
```
RealHarshThakur, line 149: I am not sure if we should be removing/managing kubeconfig context on behalf of the user. Not to mention that isn't backwards compatible. I don't know why a user would want to keep a kubeconfig of a deleted cluster in the first place. cc: @fernando-villalba

cmd/kubernetes/kubernetes_remove.go
```diff
@@ -52,53 +62,112 @@ var kubernetesRemoveCmd = &cobra.Command{
...
-				_, err = client.DeleteKubernetesCluster(v.ID)
+
+				// Delete the Kubernetes cluster
+				_, err := client.DeleteKubernetesCluster(v.ID)
 				if err != nil {
 					utility.Error("error deleting the kubernetes cluster: %s", err)
 					os.Exit(1)
 				}
+
+				/* Poll for the deletion status
```
RealHarshThakur, line 84: we should fix this on API side rather than client side so the clients don't have to poll like this.. cc : @uzaxirr

cmd/kubernetes/kubernetes.go
harsh082ip: So should I remove this flag ?
@RealHarshThakur @fernando-villalba @uzaxirr

cmd/kubernetes/kubernetes_remove.go
harsh082ip: Please correct me if I am wrong @RealHarshThakur  @uzaxirr  @fernando-villalba 

I believe the current API logic is appropriate. The API call should end after triggering the deletion of the cluster, allowing the client-side polling mechanism to take over. This approach ensures that the API call is quick and responsive. Typically, the polling mechanism requires only 1-2 attempts with a good internet connection and up to 3 attempts with a weaker connection to confirm the deletion status.

If we move the polling logic to the API side, the API call might take longer to complete and could potentially result in timeouts or errors, especially under poor network conditions. This could negatively impact user experience and increase the complexity of the API.

I look forward to your thoughts and feedback. Please let me know what I can improve.

RealHarshThakur:
Thanks for the PR @harsh082ip , we'll let you know as we discuss internally on how to proceed as some of this can be done on API side so we don't have to poll from CLI and have the user waiting.

harsh082ip:
Got it

cmd/kubernetes/kubernetes.go
fernando-villalba: > I don't know why a user would want to keep a kubeconfig of a deleted cluster in the first place

Exactly. Deleting the kubeconfig configuration automatically with a CLI tool when the cluster is deleted is extremely commong, the gcp cli does it, kind does it. It's extremely convenient and it just makes sense to do so.

cmd/kubernetes/kubernetes_remove.go
```diff
@@ -52,53 +62,112 @@ var kubernetesRemoveCmd = &cobra.Command{
...
+							utility.Error("Error deleting firewall %s: %v", firewall.Name, err)
+						} else {
+							fmt.Fprintf(os.Stderr, "Firewall %s deleted. If you want to keep the firewall in the future, please use '--keep-firewalls' flag.\n", firewall.ID)
+						}
+					}
+				}
+
+				// Delete kubeconfig if --keep-kubeconfig flag is not set
+				if !keepKubeconfig {
+					kubeconfigPath := fmt.Sprintf("%s/.kube/config", os.Getenv("HOME"))
```
insideable, line 152: you should not delete kubeconfig file because it may contain other clusters data even from different cloud providers.

alessandroargentieri:
Closing because stale and with conflicts. Thanks to everybody involved in this attempt at improvement, and apologize for the delay on our side.

Issue #394, opened 2024-04-26
Reconsider behaviour when providing a value lower than 500 for bucket size


fernando-villalba:
## Issue

The current behaviour when passing a value lower than 500 for bucket size is to ask this question:

```bash
Warning: The minimum size to create an object store is 500 GB. Would you like to create an object store of 500 GB? (y/n) ?
```

There are some issues with this approach

1. It doesn't respect the `--yes` flag:

```bash
% civo bucket create --yes --size 50 test-ticket 
Warning: The minimum size to create an object store is 500 GB. Would you like to create an object store of 500 GB? (y/n) ?
```
2. It should perhaps fail without asking anything.

If this command was part of a script, it would halt execution, if it fails, it's better. 

3. The help of the command needs to make it explicit that the minimum is 500, default is not the same as minimum, it should perhaps say `default/minimum`

```bash
-s, --size int                  Size of the Object store (default 500)
```

## Acceptance criteria

There are two ways to tackle this

* Fail lower values with a message that the minimum is 500, also change the help to explicitly say that the minimum is 500 --> preferred
* Respect the `--yes` flag, and also change the help to explicitly say the minimum is 500
* Automatically ignore lower values and assign 500 GBs, provide a warning when doing so --> This may give users the illusion they are paying less than they will if they are not paying attention, so perhaps not the best course of action.

harsh082ip:
Is there anyone working on this issue?
@fernando-villalba @uzaxirr

harsh082ip:
If not, can you please assign it to me, I want to work on this

harsh082ip:
I have made a PR regarding this issue.
Please review and let me know if any changes are needed.
@fernando-villalba @uzaxirr @RealHarshThakur @haardikdharma10

PR #438, opened 2024-07-09
fix: reconsider behavior when providing a value lower than 500 for bu…

harsh082ip:
As mentioned in issue #394, the following changes have been made:

The command will now fail immediately for values less than 500 GB with an appropriate error message.

Introduced a --yes flag to automatically adjust the bucket size to the nearest multiple of 500 without prompting the user. 
For example:
Warning: The size to create an object store must be a multiple of 500. Would you like to create an object store of 1000 GB instead? (y/n)?
Using the --yes flag, users can bypass this prompt.

Improved the help message to make the minimum and default size requirements explicit:
  -s, --size int                  Size of the Object store in GB (minimum: 500)  (default 500)

Screenshot:
image

Please review and let me know if any changes are needed.
@haardikdharma10 @uzaxirr @RealHarshThakur @alejandrojnm


harsh082ip:
Got it

alessandroargentieri:
@fulviodenza are you free to give it a look?

giornetta:
Hi @harsh082ip, thanks for your contribution!

While the original issue has already been marked as closed and some parts of it were resolved, I think we should discuss whether we want to let the `--yes` flag automatically select an higher size in this context.
This could cause some confusion among customers and some churn when they receive an higher bill than they expected.

 @alessandroargentieri @fulviodenza @uzaxirr thoughts?

cmd/objectstore/objectstore.go
```diff
@@ -45,10 +45,11 @@ func init() {
 	ObjectStoreCmd.AddCommand(objectStoreCredentialCmd)
 
 	//Flags for create cmd
-	objectStoreCreateCmd.Flags().Int64VarP(&bucketSize, "size", "s", 500, "Size of the Object store")
+	objectStoreCreateCmd.Flags().Int64VarP(&bucketSize, "size", "s", 500, "Size of the Object store in GB (minimum: 500)")
```
fulviodenza: let's use a variable for this `500` value,

cmd/objectstore/objectstore_create.go
```diff
@@ -61,21 +61,22 @@ var objectStoreCreateCmd = &cobra.Command{
...
-				os.Exit(1)
+			if !yesFlag {
+				utility.YellowConfirm("The size to create an object store must be a multiple of 500. Would you like to create an %s of %d GB instead? (y/n) ? ", utility.Green("object store"), bucketSize+(500-bucketSize%500))
+				accept, err := utility.UserAccepts(os.Stdin)
+				if err != nil {
+					utility.Error("Unable to parse the input: %s", err)
+					os.Exit(1)
+				}
+				if !accept {
+					os.Exit(0)
```
fulviodenza: this indicates success, the operation, when not accepted is not a success, we need `os.Exit(1)` for this

harsh082ip:
Hii @giornetta ,

Thanks for the context! I agree, auto-selecting a higher size with --yes might lead to unexpected charges. Happy to align on the best approach.

cmd/objectstore/objectstore_create.go
```diff
@@ -61,21 +61,22 @@ var objectStoreCreateCmd = &cobra.Command{
...
+			utility.Error("The minimum size to create an object store is 500 GB. Please provide a valid size.")
+			os.Exit(1)
 		} else if bucketSize%500 != 0 {
-			utility.YellowConfirm("The size to create an object store must be a multiple of 500. Would you like to create an %s of %d GB instead? (y/n) ? ", utility.Green("object store"), bucketSize+(500-bucketSize%500))
-			_, err := utility.UserAccepts(os.Stdin)
-			if err != nil {
-				utility.Error("Unable to parse the input: %s", err)
-				os.Exit(1)
+			if !yesFlag {
+				utility.YellowConfirm("The size to create an object store must be a multiple of 500. Would you like to create an %s of %d GB instead? (y/n) ? ", utility.Green("object store"), bucketSize+(500-bucketSize%500))
```
fulviodenza: we should specify that 500 and 1000GB are the only allowed sizes

fulviodenza (changes requested):

cmd/objectstore/objectstore_create.go
fulviodenza: `bucketSize+(500-bucketSize%500))` could be a small function

harsh082ip:
Hi @fulviodenza, apologies for the delay — I've pushed the required changes. Let me know if anything else is needed.

cmd/objectstore/objectstore.go
```diff
@@ -45,10 +50,11 @@ func init() {
 	ObjectStoreCmd.AddCommand(objectStoreCredentialCmd)
 
 	//Flags for create cmd
-	objectStoreCreateCmd.Flags().Int64VarP(&bucketSize, "size", "s", 500, "Size of the Object store (Minimum size is 500GB)")
+	objectStoreCreateCmd.Flags().Int64VarP(&bucketSize, "size", "s", minBucketSizeGB, fmt.Sprintf("Size of the Object store (Allowed sizes: %dGB or 1000GB)", minBucketSizeGB))
```
fulviodenza, line 53: why the 500 is set as constant while the 1000 isn't?

cmd/objectstore/objectstore_create.go
```diff
@@ -13,16 +13,19 @@ import (
...
 var waitOS bool
+var yesFlag bool
 
 var objectStoreCreateCmd = &cobra.Command{
 	Use:     "create",
 	Aliases: []string{"new", "add"},
 	Example: "civo objectstore create OBJECTSTORE_NAME --size SIZE",
 	Short:   "Create a new Object Store",
-	Long:    "Bucket size should be in Gigabytes (GB) and must be a multiple of 500, starting from 500.\n",
+	Long:    "Bucket size should be in Gigabytes (GB) and must be either 500 or 1000.\n",
```
fulviodenza, line 28: couldn't we use the constants of before, declared in a common package?

giornetta:
Closing as stale, no activity on this since June of last year.

PR #439, opened 2024-07-10
Change instance config to use the correct Reserved IP parameter

haardikdharma10:
## Description

The CLI does not honour the `--publicip` flag during instance create. This PR fixes that.

### DIFF
--- a/cmd/instance/instance.go
+++ b/cmd/instance/instance.go
@@ -54,6 +54,7 @@ func init() {
 	instanceCreateCmd.Flags().StringVarP(&firewall, "firewall", "l", "", "the instance's firewall you can use the Name or the ID")
 	instanceCreateCmd.Flags().StringVarP(&tags, "tags", "g", "", "the instance's tags")
 	instanceCreateCmd.Flags().StringVarP(&privateIPv4, "private_ipv4", "", "", "Private IPv4 address")
+	instanceCreateCmd.Flags().StringVarP(&reservedIPv4, "reservedip", "", "", "Reserved IPv4 address")
 	instanceCreateCmd.Flags().StringVarP(&tags, "region", "e", "", "the region code identifier to have your instance built in")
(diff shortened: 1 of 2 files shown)


cmd/instance/instance_create.go
```diff
@@ -139,8 +139,9 @@ If you wish to use a custom format, the available fields are:
 			}
 		}
 
+		// Set public ipv4 if provided
 		if publicip != "" {
-			config.PublicIPRequired = publicip
+			config.ReservedIPv4 = publicip
```
alessandroargentieri: PublicIPRequired is either true o false and it is random and at no cost. It differs from the reserved IP which costs 3 dollars a month.

haardikdharma10: Yes we set that to "true" by default - https://github.com/civo/civogo/blob/master/instance.go#L190

cmd/instance/instance_create.go
```diff
@@ -139,8 +139,9 @@ If you wish to use a custom format, the available fields are:
 			}
 		}
 
+		// Set public ipv4 if provided
```
alessandroargentieri, line 142:
```
    // Set reserved_ipv4 if provided
    if reservedIPv4 != "" {
         config.ReservedIPv4 = reservedIPv4
    }
```

haardikdharma10: Updated MR

cmd/instance/instance.go
```diff
@@ -54,6 +54,7 @@ func init() {
 	instanceCreateCmd.Flags().StringVarP(&firewall, "firewall", "l", "", "the instance's firewall you can use the Name or the ID")
 	instanceCreateCmd.Flags().StringVarP(&tags, "tags", "g", "", "the instance's tags")
 	instanceCreateCmd.Flags().StringVarP(&privateIPv4, "private_ipv4", "", "", "Private IPv4 address")
+	instanceCreateCmd.Flags().StringVarP(&reservedIPv4, "reservedip", "", "", "Reserved IPv4 address")
```
alessandroargentieri, line 57: "reserved_ipv4" is the correct param to be passed to the api-go.
do we wanna map it as `--reserved-ipv4`?

alessandroargentieri (approved):
LGTM

Issue #436, opened 2024-07-08
Wrong region displayed in message when no resources found


fernando-villalba:
## Issue

When you run the following command:

```
 civo instance --region nyc1 ls
```

And no resources are found, the wrong region will display in the message, here are examples when default region set to LON1

<img width="691" alt="image" src="https://github.com/civo/cli/assets/15110153/7ec4b91b-b978-4d69-8243-c5dc5070044d">

This is because when [issue 404](https://github.com/civo/cli/issues/404) was implemented the region output is set to show the default as shown here:

![image](https://github.com/civo/cli/assets/15110153/7ca6c2ea-3af0-4c45-a9ab-12e0abe38cbb)

## Acceptance criteria

* Ensure that the correct region shows on the output message when listing resources in a region where there are no resources.

Praveen005:
Hi Team, 

I just looked into this issue and to solve it,

we can write a check in [FinishAndPrintOutput()](https://github.com/civo/cli/blob/1b4aa90428cf9b803d00be5d544fab4399e83fad/utility/output_writer.go#L287C25-L287C45) like below, as using ` civo instance --region nyc1 ls` command will set `common.RegionSet` to `nyc1` 

```Go
       if len(ow.Values) == 0 {
                var region string
                if common.RegionSet != ""{
                    region = common.RegionSet
                }else{
		    region = config.Current.Meta.DefaultRegion
		}
		fmt.Fprintf(os.Stderr, "No resources found in region %s. For a list of regions use the command 'civo region ls'\n", region)
	}
```

I kindly request your review.

Regards,

Praveen

haardikdharma10:
That looks good to me @Praveen005. Do you want to open a PR with that change along with a screenshot that the command works and i'd be happy to approve and merge.

Praveen005:
Hey, @haardikdharma10 , 

when I run the command it doesn't go through, since I don't have an active civo account. Can you kindly test the changes on your end?

<img width="750" alt="Screenshot 2024-07-11 125143" src="https://github.com/civo/cli/assets/42026111/4a9c1ae7-f4a3-489a-bb44-f864f401bd2e">


Thank you.

haardikdharma10:
@Praveen005 - I tested the change and it works as expected. I have merged the PR and closing this issue. The change should be visible when we release our next version.

PR #440, opened 2024-07-11
fix: display correct region in message when no resources are found

Praveen005:
## Description

This PR addresses the issue in #436. It ensures the correct region is displayed when no resources are found.

## Changes

- Added a check to determine if the user has specified a region via the command line flag `--region`.
- If a region is specified, its value is used.
- If no region is specified, the default region is printed.


haardikdharma10 (approved):
LGTM! Thanks @Praveen005 :)

Praveen005:
Thanks for the review and approval, @haardikdharma10

Issue #392, opened 2024-04-26
Reconsider CLI defaults


fernando-villalba:
## Issue

Creating a cluster with `civo kubernetes create` creates a cluster with defaults. This is phenomenal use of convention over configuration so the user doesn't need to input too many flags or commands to create a cluster.

However we should perhaps reconsider the defaults used here, right now this is what it does:

1. Create a cluster with 3 `g4s.kube.medium` nodes --> We should probably make the `g4s.kube.xsmall` the default. 3 nodes may be okay, although perhaps 2 is more conservative.
1. Randomly pick a name for the cluster --> This is phenomenal, no need to change
1. Randomly pick a Linux distro for cluster --> I don't think this is a good thing as it changes the nature of the cluster too much. 
1. Create a firewall/s for the cluster --> This is also okay, but cluster deletion should clean these, or at least provide an option to do so. Will create another ticket for this. 

(Not sure if I am missing other defaults)

In addition, the UI should be aligned with the defaults the CLI provides. 

## Acceptance Criteria

 1. Reconsider changing the defaults to create a cluster that's as cheap as possible for generalised workloads.
 1. Make the OS deterministic, don't randomize it. 

Cluster defaults should be

* As secure as possible.
* As cheap as possible.
* Catering to general needs of the public wishing to test a cluster or use for quick, simple workloads. We perhaps shouldn't assume that the defaults are for heavy workloads, or even for production workloads as in those scenarios a user may want more fine grained settings
* The nature of the cluster should be predictable, randomizing the OS prevents that.

Another option could be to create templates for different types of configs, for example:

* ML
* Production
* Test
* Budget
* Etc

But this is out of the scope of this ticket.

harsh082ip:
Hi Team,

I have addressed several aspects of this issue:

Set the default node size to g4s.kube.xsmall
Set the default number of nodes to 2
Ensured firewall rules are correctly applied during cluster creation
However, I was unable to find a way to make the OS selection deterministic. There doesn't seem to be any mention of this in the API documentation. Could you provide guidance on how to achieve this?

Regards,
Harsh

harsh082ip:
I have made a PR regarding this issue, 
Please review it 
@haardikdharma10 @fernando-villalba

harsh082ip:
#441

harsh082ip:
Please refer this 👆👆

PR #441, opened 2024-07-11
fix: Reconsider CLI defaults #392

harsh082ip:
As discussed in issue-392, I have implemented the following fixes:

Set the default node size to g4s.kube.xsmall
Set the default number of nodes to 2
Ensured firewall rules are correctly applied during cluster creation
However, I was unable to make the OS selection deterministic. The API documentation does not seem to provide any guidance on this. Could you please advise on how to achieve this?

Screenshots:
image
Screenshot from 2024-07-11 15-20-10

Please review and let me know if any further changes are needed.

@haardikdharma10 @RealHarshThakur @uzaxirr @fernando-villalba


alessandroargentieri:
we have set different defaults months ago. thanks for contributing and sorry for the delay in the response

Issue #444, opened 2024-07-17
The CLI should support updating the firewall of a cluster


fernando-villalba:
## Issue

Currently there is no way (that I know of) to change the firewall of a cluster, this is supported by the API, hence it should be possible to do in the CLI as well. 

## Acceptance criteria

* Support updating the firewall id of a cluster.

haardikdharma10:
Fixed with https://github.com/civo/cli/pull/445

PR #445, opened 2024-07-17
Add ability to update cluster firewall

haardikdharma10:
## Description

This PR- 
- Adds a new command to allow updating firewall for a kubernetes cluster.
- Bumps `civogo` module to the latest version.

## Fixes
https://github.com/civo/cli/issues/444

### DIFF
--- a/cmd/kubernetes/kubernetes_update.go
+++ b/cmd/kubernetes/kubernetes_update.go
@@ -1,0 +1,72 @@
+package kubernetes
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/civo/civogo"
+	"github.com/civo/cli/common"
+	"github.com/civo/cli/config"
+	"github.com/civo/cli/utility"
+	"github.com/spf13/cobra"
+)
+
+var firewall string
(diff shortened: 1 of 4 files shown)


cmd/kubernetes/kubernetes_update.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/civo/cli/utility"
+	"github.com/spf13/cobra"
+)
+
+var firewall string
+
+var kubernetesUpdateCmd = &cobra.Command{
+	Use:     "update",
+	Aliases: []string{"edit"},
+	Short:   "Update a Kubernetes cluster",
```
fernando-villalba: You may want to be more specific here  "Update the configuration of your Civo Kubernetes cluster" or something like that --> This makes it more specific to the user that they are updating just the configuration platform side, if that makes sense.

haardikdharma10: updated

fernando-villalba (approved):
LGTM!

PR #459, opened 2024-09-04
Allow scaling from 0 to one or more nodes

fulviodenza:
As it was before, we were taking from the Pools and not from RequiredPools the node to update, this was causing the scaling of the nodes from 0 to 1 (or more) to fail and was deleting the pool.


alessandroargentieri (approved):
LGTM

Issue #390, opened 2024-04-25
Creating a kubernetes cluster should configure kubectl by default and clean the configuration on deletion


fernando-villalba:
## Issue

When you create a cluster with tools like `kind` or `gcloud` the default behaviour is to create the cluster, add the configuration to `kubectl` and then switch the context to that cluster. 

Once the cluster is deleted, then the configuration for that cluster is also deleted from `kubectl`

Civo CLI has multiple issues in this regard.

### 1. `--save` and `--wait` to add configuration 

Using the command:

```bash
civo kubernetes create --save --wait
```
Prompts this question to the user:

```bash
Warning: Are you sure you want to overwrite the kubernetes config (y/N) ?
```

There are very few situations where a user would want to completely overwrite their kubectl configuration, if done by accident this could be a very tedious issue to solve, especially if the user had multiple working configurations. Funnily enough the command `civo cluster configuration` offers an `--overwrite` flag for this which is a lot more sensible, the default behaviour for `--save` is to merge.

### 2. Using `--save`, `--wait`, `--merge`, `--switch`

Doing:

```bash
civo kubernetes create --save --wait --merge --switch
```

Provides the behaviour a user would be used to with other kubernetes cluster creation tools

However the problem is that when you delete a cluster created with these flags, it does not delete the configuration. That means the users are left with a lot of dead and useless configuration in their `~/.kube/config` folder that they have to clean. 

### 1. When using flags that require a bunch of other flags, make their use implicit.

For example, if the user uses `--switch`, don't force them to also specify `--wait` and `--merge` and `--save`, instead, implicitly assume that's the case, and the same for all others. `--merge` assuming `--save` and so on...

### 2. Automatically delete the kubectl config once the cluster is deleted.

This should be whether the config is created on cluster creation or added with `civo kubernetes config`. The CLI should delete the configuration by default. A flag can be provided to prevent the deletion of the configuration if desired. 

### 3.  Don't overwrite the config with `--save`

Fine to have an `--overwrite` flag, but don't do it by default. Always merge by default.

### 4. Discuss whether this should be the default behaviour or not.

We should discuss whether we would want this to be the default behaviour or not. Right now the CLI creates clusters assuming that the user is creating them in batches of multiple clusters where the configuration may not be desired, whereas tools like `kind` assume that the user wants to use the cluster straight after - perhaps this would be the most common use case for us or not, in which case it should be the default behaviour.

anujagrawal380:
[...]

PR #463, opened 2024-09-09
rm kubeconfig after deleting cluster

uzaxirr:
Partially solves: #390 

Screenshot 2024-09-09 at 2 46 27 PM


alessandroargentieri:
@uzaxirr what does "partially solves" mean in this context? What is left out from the solution you proposed?

uzaxirr:
> @uzaxirr what does "partially solves" mean in this context? What is left out from the solution you proposed?

Means that it only has keep-config flag support as of now. `-wait` , `--merge` and `--save` are still not implemented. Need to re-review this PR.

PR #470, opened 2024-09-23
Change of behaviour when bucket size is lower than 500gb

### LINKED ISSUE #394: Reconsider behaviour when providing a value lower than 500 for bucket size

uzaxirr:
fixes: #394

Screenshot 2024-09-24 at 1 35 43 AM

### DIFF
--- a/cmd/kubernetes/kubernetes.go
+++ b/cmd/kubernetes/kubernetes.go
@@ -100,7 +100,6 @@ func init() {
 	kubernetesCreateCmd.Flags().StringVarP(&clusterType, "cluster-type", "", "k3s", "optional, possible options: k3s,talos.")
 
 	kubernetesRenameCmd.Flags().StringVarP(&kubernetesNewName, "name", "n", "", "the new name for the cluster.")
-	kubernetesRemoveCmd.Flags().BoolVar(&keepConfig, "keep-config", false, "Keep the kubectl config after deleting the cluster")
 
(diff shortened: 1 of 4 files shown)


vishalanarase (approved):
LGTM

PR #476, opened 2024-10-09
Add support for attached volumes in instance_show.go

uzaxirr:
Screenshot 2024-10-09 at 11 50 40 AM


fernando-villalba:
We need to be consistent here. You have `Network ID, Disk Image ID, etc` and then you have `Attached Volumes`

Maybe change it to `Attached Volumes IDs`

fernando-villalba (approved):

Issue #477, opened 2024-10-11
Error when trying to run  "apikeys" commands after initially adding a random API key.


olamilekan000:
## Issue

I  encounter an error when trying to run any "apikeys" sub-commands like, `remove`, `list`, `show`, and `save` after successfully adding an API key for the first time. The error is as follows:

```
Unable to list regions: DatabaseAccountNotFoundError: Failed to find the account within the internal databaseError getting supported regions to feature DatabaseAccountNotFoundError: Failed to find the account within the internal database 
```

When I checked why this error occurs, I discovered it was due to this block
```
	if Current.Meta.CurrentAPIKey != "" && Current.RegionToFeatures == nil {
		Current.RegionToFeatures, err = regionsToFeature()
		if err != nil {
			fmt.Printf("Error getting supported regions to feature %s \n", err)
			os.Exit(1)
		}

		dataBytes, err := json.Marshal(Current)
		if err != nil {
			fmt.Printf("Error parsing JSON %s \n", err)
			os.Exit(1)
		}

		err = os.WriteFile(filename, dataBytes, 0600)
		if err != nil {
			fmt.Println(err)
			os.Exit(1)
		}
	}
```
The API key is set, but it fails when trying to fetch the features for the default region using that key."

 ## Acceptance Criteria

- I should be able to run the `save`  subcommand as many times as needed without errors.
- I should be able to run the `remove`  subcommand as many times as needed without errors.
-  I should be able to run the `list`  subcommand as many times as needed without errors.
-  I should be able to run the `remove`  subcommand as many times as needed without errors.

uzaxirr:
Should be fixed via https://github.com/civo/cli/pull/486

PR #478, opened 2024-10-11
fix error when trying to run apikey sub-commands

olamilekan000:
```
change fixes error when trying to run `apikey` sub-commands after successfully saving an API key.

```

Reported issue can be found [here](https://github.com/civo/cli/issues/477).


giornetta:
Hi, the issue you linked has been marked as closed, thus I will close this MR as well.

Thanks for your contribution though, and sorry for the late reply!

PR #481, opened 2024-10-17
fix: display correct cluster type during creation with --wait flag

Praveen005:
When you create a cluster with `--wait` flag, the spinner message says, creating a `k3s` cluster, even if the cluster type is `talos`, now it takes the value from from `--cluster-type` flag, or defaults to `k3s` if nothing specified.

**Issue:**

If you see, we are creating a `talos` cluster, but the spinner message says `k3s`:

Screenshot 2024-10-17 202844

------------------------------------------

**Result:**

Correctly picking up the cluster type now:

Screenshot 2024-10-17 213207

Rightly defaulting to `k3s`:

Screenshot 2024-10-18 013946

### DIFF
--- a/cmd/kubernetes/kubernetes_create.go
+++ b/cmd/kubernetes/kubernetes_create.go
@@ -225,7 +225,7 @@ var kubernetesCreateCmd = &cobra.Command{
 
 			stillCreating := true
 			s := spinner.New(spinner.CharSets[9], 100*time.Millisecond)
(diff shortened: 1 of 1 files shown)


fulviodenza:
Closing as this has been fixed here https://github.com/yuvarajrece/cli/commit/f0887c5100f888661ebf47635864ce7c9e3d3fc5

PR #488, opened 2025-01-06
create instance tests

uzaxirr:


alessandroargentieri:
@uzaxirr do we close it or can you move from it from Draft? Thanks

cmd/instance/instance_create_test.go
```diff
@@ -0,0 +1,113 @@
+package instance_test
```
fulviodenza, line 1: there is a lot of replicated code here, everything in a for loop would be better

fulviodenza (changes requested):

uzaxirr:
We need to close it for now

PR #548, opened 2025-05-11
feat: add CLI commands to status and stop instance vnc sessions

Faizan-Alam-1:
## feat: add CLI commands to manage instance console sessions

This update introduces new CLI commands to interact with instance VNC console sessions.

#### Added

- `civo instance console status <instance_id>`  
  Checks if a VNC session is currently active for the given instance.  

status

-  `civo instance console stop <instance_id>`  
  Terminates an active VNC session before its expiration.  
  
stop

#### Region Handling

- If the `--region` flag is not specified, the default region will be used.

#### Examples

civo instance console status <instance_id> --region <region_name>

Screenshot from 2025-05-12 01-28-53

civo instance console stop <instance_id> --region <region_name>

Screenshot from 2025-05-12 01-29-25

WhatsAppVideo2025-05-12at1 08 58AM-ezgif com-video-to-gif-converter

### DIFF
--- a/cmd/instance/instance_vnc.go
+++ b/cmd/instance/instance_vnc.go
@@ -123,3 +123,7 @@ func waitEndpointReady(url string) error {
 		time.Sleep(7 * time.Second) // Wait for 7 seconds before the next attempt
 	}
 }
+func init() {
+	instanceVncCmd.AddCommand(statusCmd)
+	instanceVncCmd.AddCommand(stopCmd)
+}
--- a/cmd/instance/instance_vnc_status.go
+++ b/cmd/instance/instance_vnc_status.go
@@ -1,0 +1,93 @@
+package instance
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/civo/cli/config"
+	"github.com/civo/cli/utility"
+	"github.com/spf13/cobra"
+)
+
+type ActiveVNCResponse struct {
+	URI        string `json:"uri"`
+	Expiration string `json:"expiration"`
+}
+
+type InactiveVNCResponse struct {
+	Code   string `json:"code"`
+	Reason string `json:"reason"`
+}
+
+func GetVNCSession(baseURL, region, token, instanceinstanceID string) {
+	url := fmt.Sprintf("%s/v2/instances/%s/vnc?region=%s", baseURL, instanceinstanceID, region)
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		utility.Error("Request creation error:%s", err)
+		return
+	}
+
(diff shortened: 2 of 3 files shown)


giornetta:
Hi, thanks for picking this up. However this approach is not what we usually follow for the CLI: HTTP requests are not defined here, but in the https://github.com/civo/civogo Go SDK. The CLI uses that to communicate with the Civo API.

Faizan-Alam-1:
@giornetta Thanks for the feedback. I now understand that I should use the Civogo SDK instead of making my own HTTP requests. I’ll update my code to follow that approach. I just want to try again and do it the right way this time.   I also noticed that the API supports deleting a VNC session using this URL: **DELETE /v2/instances/{id}/vnc?region={region}** —but the SDK doesn’t have a method related  to this .  If you can help me add that, it would be really helpful for me.

giornetta:
@Faizan-Alam-1 Yes, the task would also involve adding the method to Civogo. This is a fairly simple task in this case, so you're free to tackle it if you want (even though @uzaxirr was already going to work on that himself!), but generally I'd suggest you pick up tasks that do not need to implement new features as only the Civo team has access to our internal API documentation, and it might be difficult for external contributors to gather all the context required to implement new features.

Faizan-Alam-1:
@giornetta Can I work on adding the **civo instance console status <instance-id>** command? The **GetInstanceVNC()** function returns the VNC duration and is already implemented, so I won’t need any new API for it.

giornetta:
@Faizan-Alam-1 The `GetInstanceVNC` function you mentioned is not the correct one for the `console status` command.

That function sends a `PUT` request to create a new session. It was named like that because we didn't have a GET/DELETE endpoint at the time. But as I said new methods need to be implemented in civogo.

uzaxirr:
@Faizan-Alam-1 i would suggest you to take a look at https://github.com/civo/civogo/
how everything works is that the CLI is just a wrapper around civogo SDK. The actual API calls takes place in the SDK not in the CLI.
Closing this PR for now

Faizan-Alam-1:
@uzaxirr @giornetta  Hey, I implement the GetInstanceVncStatus () in civogo sdk, i just implemented  it locally, not I have push yet. please just check it out. done correctly ? 

This code SS 

Screenshot from 2025-05-18 01-23-22


API response when vnc session  active

Screenshot from 2025-05-18 01-18-31

api response when vnc session not active

Screenshot from 2025-05-18 01-22-33

also, i have adding giti file so that you check it out.

https://gist.github.com/Faizan-Alam-1


If it is correct then i will make pull request and also implement for delete request too, please review it.  I will also write the test case  for this function  if you want .

uzaxirr:
This seems correct, you just don't  have to explicitly add the region param in the URL. It is configured in the client struct. Please raise a PR

PR #557, opened 2025-07-29
Add /scheduled as backupID if `--scheduled` flag is found

fulviodenza:
We need to specify, as backupID the string "scheduled" as Databases can only have one scheduled backup at a time and they don't have an ID since they're bound to the CivoDatabase resource


alessandroargentieri (approved):
LGTM

PR #558, opened 2025-08-11
fix: upgrade civogo dependency to fix bug in instance selection

giornetta:
Instance selection would fail when instances with a similar name exist. This was caused by a bug in civogo. This MR updates the Civogo version used by CLI to bring in the bug fix.


alessandroargentieri (approved):
LGTM

PR #561, opened 2025-09-07
feat: add custom api url to search against

jokestax:
Added an extra field called APIURL, making it easier to test against custom api-go endpoints,where apikey save will now prompt for an extra field APIURL, where you can enter https://api-staging.civo.com you then run ./civo apikey show

image


giornetta:
Do we want this? This is mostly useful for us Civo staff with access to different environments, but for customers there's no value in adding an extra step to configure their API Key in my opinion.

jokestax:
thanks @giornetta ,i will close this for now

Issue #563, opened 2025-10-17
objectstore used stats


NerdyShawn:
The show function of an objectstore has **Stats** of a bucket but the `Size Used` value does not display a unit making it too ambiguous. 

```bash
civo version && civo objectstore show somecoolbucket01 | grep Stats
Civo CLI v1.4.3
                Stats : Objects: 34, Size: 500 MB, Size Used: 12356
```

Proposing we display a unit on the returned value for size used.

fulviodenza:
Hi @NerdyShawn thanks for raising this. I created a possible fix as you can see from the above activity. 
That is my solution to disabiguate the value and have MB instead of KB. If you have some suggestion/feedback/idea please let me know.

NerdyShawn:
left some comments on the PR, also I just noticed today the **Size** value unit is just wrong too, the objectstore is `500 GB` not MB. The smallest size an objectStore can be provisioned is currently 500GB.

giornetta:
Merged in #564, will be released in `v1.4.7`.

PR #564, opened 2025-10-17
Add MB to used objectstore output

fulviodenza:
Disambiguating value adding "KB" to the output of civoobjectstore show command #563


NerdyShawn (commented):
Nice I think displaying some unit is better than no unit for sure. The only other thing I was thinking was making it a bit dynamic something like a `df -h` where its relative to the size consumed. I haven't tried this below but the 🤖  vibes was suggesting something like that to get dynamic, without needing an additional dependency. If this is a bad idea, feel free to say its too much. 😃 

```go
// humanSize returns a human‑readable string for a value in bytes
// using powers of 1024: B, KB, MB, GB, TB, PB.
func humanSize(bytes int64) string {
    const (
        _          = iota
        KB float64 = 1 << (10 * iota)
        MB
        GB
        TB
        PB
    )
    // use float64 so we can keep two decimal places
    value := float64(bytes)

    switch {
    case value >= PB:
        return fmt.Sprintf("%.2f PB", value/PB)
    case value >= TB:
        return fmt.Sprintf("%.2f TB", value/TB)
    case value >= GB:
        return fmt.Sprintf("%.2f GB", value/GB)
    case value >= MB:
        return fmt.Sprintf("%.2f MB", value/MB)
    case value >= KB:
        return fmt.Sprintf("%.2f KB", value/KB)
    default:
        return fmt.Sprintf("%d B", bytes)
    }
}
```
then update the variable reference if we think this is a good idea

```go
// sizeUsedMB := float64(stats.SizeKBUtilised) / 1024.0
// ow.AppendDataWithLabel("stats", fmt.Sprintf("Objects: %d, Size: %d MB, Size Used: %.2f MB", stats.NumObjects, objectStore.MaxSize, sizeUsedMB), "Stats")

// ---- NEW ----
sizeUsedBytes := int64(stats.SizeKBUtilised) * 1024
humanSizeUsed := humanSize(sizeUsedBytes)

ow.AppendDataWithLabel(
    "stats",
    fmt.Sprintf("Objects: %d, Size: %d MB, Size Used: %s", stats.NumObjects, objectStore.MaxSize, humanSizeUsed),
    "Stats",
)
```

fulviodenza:
Sure thing @NerdyShawn, I agree with the suggestions, I will update the code soon

NerdyShawn:
Sounds good thanks! Also left a [comment on the issue ](https://github.com/civo/cli/issues/563#issuecomment-3443875263)but the unit on the 500 is also wrong, that should be a static  500GB I believe.

cmd/objectstore/objectstore_show.go
```diff
@@ -79,3 +86,30 @@ var objectStoreShowCmd = &cobra.Command{
 		}
 	},
 }
+
+func humanSize(bytes int64) string {
```
giornetta, line 90: nit (sorry): worth adding unit tests for this?

cmd/objectstore/objectstore_show.go
```diff
@@ -66,7 +66,14 @@ var objectStoreShowCmd = &cobra.Command{
 		ow.AppendDataWithLabel("region", client.Region, "Region")
 		ow.AppendDataWithLabel("accesskey", creds.AccessKeyID, "Access Key")
 		ow.AppendDataWithLabel("status", objectStore.Status, "Status")
-		ow.AppendDataWithLabel("stats", fmt.Sprintf("Objects: %d, Size: %s MB, Size Used: %d", stats.NumObjects, strconv.Itoa(objectStore.MaxSize), stats.SizeKBUtilised), "Stats")
+		sizeUsedBytes := int64(stats.SizeKBUtilised) * 1024
+		humanSizeUsed := humanSize(sizeUsedBytes)
+
+		ow.AppendDataWithLabel(
+			"stats",
+			fmt.Sprintf("Objects: %d, Size: %d MB, Size Used: %s", stats.NumObjects, objectStore.MaxSize, humanSizeUsed),
```
giornetta: `Size: %d MB` should be `Size: %d GB`

giornetta: I would have:
`"Objects: %d, Size: %d GB, Used: %s"`

giornetta (approved):

PR #573, opened 2025-11-25
fix: pluralizations of the verb "to have" across the repository

giornetta:
When deleting multiple resources, we're pluralizing the resource type in case multiple resources were deleted at the same time, so that the CLI displays the correct word to the user.

However, the rest of the phrases were not being pluralized correctly.

This fixes it by internalizing the dependency on `github.com/alejandrojnm/go-pluralize` and extending it by adding a utility `pluralize.Has` function that uses `pluralize.Pluralize` to correctly display "has" or "have" depending on the size argument passed in.

This new utility has been added to all cases where this problem could be found (mostly when deleting multiple resources at once).


dippydocus (approved):

fulviodenza (approved):
LGTM!

PR #579, opened 2026-03-03
feat: add civo vpc command suite with full CRUD for VPC resources

aculnaig:
Implement the complete `civo vpc` command hierarchy using the VPC-prefixed methods from civogo SDK v0.7.0-alpha1. All commands target the /v2/vpc/ API path and follow existing CLI patterns for output formatting, batch deletion, region handling, and flag conventions.

Command tree:

  civo vpc network      ls | create | show | update | remove
  civo vpc subnet       ls | create | show | remove | attach | detach
  civo vpc firewall     ls | create | show | update | remove
  civo vpc firewall rule  ls | create | remove
  civo vpc loadbalancer ls | create | show | update | remove
  civo vpc ip           ls | reserve | show | update | assign | unassign | delete

Implementation details:
- 37 new files in cmd/vpc/ covering network, subnet, firewall, firewall rules, load balancer, and reserved IP resources
- Wire VPCCmd into rootCmd via cmd/root.go
- Subnet commands require --network flag to scope operations to a network
- Load balancer create/update support repeatable --backend flag with format ip:source_port:target_port:protocol:health_check_port
- IP assign uses --resource-id and --resource-type for generic resource assignment (not limited to instances)
- Firewall list resolves network IDs to names via ListVPCNetworks
- Firewall show includes inline rule listing
- All remove commands support batch deletion with confirmation prompts
- Three-way output support: json, custom fields, and human-readable

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v6
         with:
-          go-version: ${{ vars.GO_VERSION || '1.24.11' }}
+          go-version: ${{ vars.GO_VERSION || '1.24.13' }}
           cache: true
 
       - name: Verify dependencies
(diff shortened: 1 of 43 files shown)


giornetta:
Is this something we want? Aren't we making our CLI API more complicated?

giornetta (approved):

PR #585, opened 2026-04-22
fix(firewall): do not require --startport for ICMP rules

### LINKED ISSUE #350: Create Firewall Rule for ICMP Requires Starting Port Number
Environment: Windows 10 x64
Version: Civo CLI v1.0.69
Bug Type: Minor
Priority: Low
Workaround Possible: Yes
Summary: requires a command flag that should not be specified

Example command:
 civo firewall rule create %NAME% ^
  --protocol=ICMP ^
  --direction=ingress ^
  --label="Allow all ICMP traffic" ^
  --action="allow" ^
  --cidr=1.2.3.4/32

Error message:
Error: required flag(s) "startport" not set

Temporary workaround:
Add the starting port number:
--startport=1

Bug fix objective:
Do not require specifying starting port number for ICMP rules

alubbock:
ICMP is a portless protocol, so --startport should not be mandatory when creating firewall rules with --protocol=ICMP.

Two issues were found and fixed:

1. MarkFlagRequired("startport") was enforced globally by Cobra before the command ran, making it impossible to omit regardless of protocol. Removed the declaration and replaced it with a runtime check that only errors for TCP/UDP.

2. The Civo API is case-sensitive: it only skips the port-empty validation when protocol is sent as lowercase "icmp". Sending "ICMP" triggered a 400 firewall_ports_should_not_be_empty error even though start_port and end_port were present as empty strings in the JSON body. Fixed by normalising protocol to lowercase before building the request config.

EndPort assignment and success output messages are also updated to omit port information for ICMP rules.

Fixes #350


giornetta (approved):
LGTM, thanks!

PR #586, opened 2026-04-22
fix: honour -o json/custom across missing and broken commands

alubbock:
Several commands ignored the `-o json` and `-o custom` flags, always
printing plain text regardless. One command (`kubernetes cluster remove`) 
had a latent bug where `-o json` also triggered the custom formatter.

**Missing output format handling (always printed plain text):**
  - `kubernetes node-pool instance-delete`
  - `kubernetes cluster remove` (also had `WriteCustomOutput` inside the
    `"json"` case, so `-o json` produced both outputs)
  - `apikey remove` / `apikey set`
  - `instance public-ip` / `instance recovery`
  - `kubernetes update-kubeconfig`
  - `network show`

**Also fixed in `instance recovery`:** a `fmt.Println("Instance ID is:
…")` that fired before the API call, unconditionally, on every
invocation.


cmd/network/network_show.go
```diff
@@ -29,36 +30,57 @@ var networkShowCmd = &cobra.Command{
...
+		ow := utility.NewOutputWriter()
+		ow.StartLine()
+		ow.AppendDataWithLabel("id", network.ID, "ID")
+		ow.AppendDataWithLabel("name", network.Name, "Name")
+		ow.AppendDataWithLabel("default", utility.BoolToYesNo(network.Default), "Default")
+		ow.AppendDataWithLabel("cidr", network.CIDR, "CIDR")
+		ow.AppendDataWithLabel("status", network.Status, "Status")
+		ow.AppendDataWithLabel("ipv4_enabled", utility.BoolToYesNo(network.IPv4Enabled), "IPv4 Enabled")
+		ow.AppendDataWithLabel("ipv6_enabled", utility.BoolToYesNo(network.IPv6Enabled), "IPv6 Enabled")
+		ow.AppendDataWithLabel("vlan_id", fmt.Sprintf("%d", network.VlanID), "VLAN ID")
```
giornetta: The fields `vlan_id`, `physical_interface`, `allocation_pool_v4_start` and `allocation_pool_v4_end` should only be displayed if the network is VLAN-enabled.

This way, `vlan_id` is shown in the JSON output as `"vlan_id":"0"`, whereas the other fields are silently dropped because they're empty. I would add a guard here to prevent `vlan_id` from appearing in the JSON output as it's happening in the human formatting.

cmd/network/network_show.go
```diff
@@ -29,36 +30,57 @@ var networkShowCmd = &cobra.Command{
...
-		// Nameserver Details
-		if len(network.NameserversV4) > 0 || len(network.NameserversV6) > 0 {
-			fmt.Println("\nNameserver Details:")
-			if len(network.NameserversV4) > 0 {
-				fmt.Printf("Nameservers IPv4: %s\n", utility.SliceToString(network.NameserversV4))
+			if network.VlanID != 0 {
+				fmt.Println("\nVLAN Details:")
+				fmt.Printf("VLAN ID: %d\n", network.VlanID)
+				fmt.Printf("Hardware Address: %s\n", network.PhysicalInterface)
+				fmt.Printf("Gateway IPv4: %s\n", network.GatewayIPv4)
```
giornetta: This is actually a pre-existing "bug" we could solve here: we can show the gateway IP even for non-VLAN networks, so this can be moved outside the guard.

cmd/network/network_show.go
giornetta: Also a minor nitpick, we're using `strconv.Itoa(network.VlanID)` in network_list.go, we should probably stick to that for consistency.

cmd/instance/instance_recovery.go
```diff
@@ -47,8 +47,6 @@ Example:
 			utility.Error("%s", err)
 			os.Exit(1)
 		}
-		fmt.Println("Instance ID is: ", instance.ID)
```
giornetta, line 50: By removing this line, the human-form output no longer shows the UUID of the instance, but only the hostname. Consider adding it (between parentheses?) in the human-form output, next to the hostname

cmd/network/network_show.go
alubbock: Thanks, fixed. The VLAN-specific fields are now guarded behind VlanID != 0, so they're omitted from JSON output entirely for non-VLAN networks. Also switched to strconv.Itoa for consistency with network list.

cmd/network/network_show.go
alubbock: Moved outside the guard so it appears for all networks in both JSON and human-readable output.

cmd/instance/instance_recovery.go
alubbock: Good spot - added the UUID in parentheses after the hostname.

giornetta (approved):
LGTM!

PR #594, opened 2026-06-22
fix(kubernetes): block metrics-server install on Talos clusters

### LINKED ISSUE #422: Talos kubernetes-marketplace compatibility - block install attempt for cluster-type talos

hlts2:
Fixes #422

## Problem
When adding a marketplace app to a Talos cluster, the CLI reports a successful install even though nothing is actually installed:

```
civo k8s apps add metrics-server -c talosmetrics01
The application was installed in the Kubernetes cluster talosmetrics01
```

The marketplace `metrics-server` is only enabled for k3s clusters (https://github.com/civo/kubernetes-marketplace/blob/master/metrics-server/manifest.yaml), so the user has no way to know the install silently did nothing.

## Scope
Only `metrics-server` is blocked on Talos. Other marketplace apps install fine on Talos clusters, so blocking all apps (as explored in #593) would be too broad. On Talos, metrics-server must be installed separately (https://docs.siderolabs.com/kubernetes-guides/monitoring-and-observability/deploy-metrics-server).

## Fix
- `checkAppPlan` now also returns the index of the resolved app, so validation runs against the **canonical** app name rather than raw user input. This closes the partial-match bypass (e.g. `metrics` resolves to `metrics-server` via `strings.Contains`).
- `RequestedSplit` exits early with a clear message when the resolved app is unsupported on the target cluster type.
- Added a `unsupportedAppsByClusterType` table with a TODO to remove it once `cluster_type` is surfaced through the API/civogo.

## Tests
- Table-driven tests for `isAppCompatibleWithClusterType`.

Supersedes #593.


hlts2:
### Manual verification

Confirmed the behavior against real clusters.

**Talos cluster — blocked with a clear error:**
```
❯ ./cli kubernetes applications add metrics-server --cluster hiroto-talos
Error: metrics-server cannot be installed via the marketplace on talos clusters
```

**k3s cluster — installs as before:**
```
❯ ./cli kubernetes applications add metrics-server --cluster hiroto-k3s-1-36
The application was installed in the Kubernetes cluster hiroto-k3s-1-36
```

**Talos cluster — other applications still install fine (no regression):**
```
❯ ./cli kubernetes applications add loki --cluster hiroto-talos
The application was installed in the Kubernetes cluster hiroto-talos
```
Only `metrics-server` is blocked on Talos; all other marketplace apps are unaffected.

go.mod
```diff
@@ -1,4 +1,4 @@
-go 1.26.3
+go 1.26.4
```
hlts2, line 1: I have upgraded the version to resolve the security warning.
- https://github.com/civo/cli/actions/runs/27932229210/job/82646254824

jokestax (approved):

